Issue #2, opened 2022-01-10
Plan out how you will implement this; include any new command line arguments and any new dependencies you'll use


johndgiese:

xaker00:
## Action plan
+ add get_page method to notion client (notion.py)
+ study notion api page object and come up with a way to create pandoc json from it
+ use pandoc to create markdown
+ add output location as command line argument?

## New dependencies
+ Pandoc 
	May not be strictly required, but if we have pandoc json, then generating any output is trivial, not just markdown. Better to use an existing AST than making our own.

johndgiese:
This approach looks good to me! Could you add a few details about how you plan to test the tool? In particular, what would be the inputs and outputs for the tests? Note that we have a GitHub action that automatically runs tests when you push them up. Ideally we'd figure out some way to include our tests in there.

xaker00:
It looks like there is already a GitHub action to run tests. I plan to add snippets of Notion page objects (in JSON) and compare to corresponding known good pandoc objects.

Issue #1, opened 2022-01-06
Add the ability to export a notion database into a set of markdown files


johndgiese:
A row in a Notion database can also be a document. Thus, one way to export a notion database is to create a set of markdown files---one for each row in the database. The actual data in the database can be added as YAML front matter in the markdown files.

For example, if you have a database with two columns, "title" and "description" and two rows, you could create two markdown files. Each Notion database has a single required column with the "name" type. The names of the files can be taken from the column with the "name" type, in this case the "title" column.

| title (type "name") | description (type "text") |
| --- | --- |
| Intro to Quality Management Systems | A quality management system blah blah blah |
| Software Risk Management | Risk Management blah blah blah |

This would produce two files. The first one be`Intro to Quality Management Systems.md` and would have this content:

```
---
title: Intro to Quality Management Systems
description: A quality management system blah blah blah
---

The content of the page would go here.

It would be formatted using Markdown, including lists, etc.

Images would be saved in a specified folder with relative paths from the markdown file to that folder.
```

You may want to convert the Notion block JSON into Pandoc's JSON AST format and then let Pandoc convert this AST into markdown.

- [x] #2
- [x] #3
- [x] Add support for bold, italics, and links
- [x] Add support for horizontal dividers, code blocks, and block quotes
- [x] Add support for images
- [x] Add support for tables
- [x] Plugin system for handling toggle lists
- [x] Add support for markdown in a "content" property if you are generating a single YAML file vs a set of MD files
- [x] When generating a YAML file from a database if you have rows that don't have an associated page then set the "content" property to `null`.
- [x] Handle bold, italics, and external links in text properties.
- [x] Add support for checkboxes

- In general, don't worry about links to Notion pages. (Will solve later)

johndgiese:
Any blocks we don't support print a warning to stderr and then not include the unsupported notion blocks in the generated output (but produce as much as we do support).

johndgiese:
When generating a YAML file from a database if you have rows that don't have an associated page then set the "content" property to `null`.

PR #5, opened 2022-02-03
Basic working version

xaker00:
Notion Blocks supported:
- Bookmarks
- Bulleted lists
- Child pages (pages linked from database)
- Code (fenced)
- Divider
- Headings
- Images (download embedded, link to external)
- Numbered Lists
- Paragraphs
- Quotes
- Tables
- Todo (checkboxes)
- Toggle

Supported styling:
- Bold
- Italic
- Inline code
- strikethrough

Blocks can be extended or replaced using a plugin system. 

Addresses issues #1, #2, & #3.


n2y/converter.py
```diff
@@ -0,0 +1,466 @@
...
+#   bulleted_list - Notion has bulleted_list_item, but no enclosing container
+#   numbered_list - Notion has numbered_list_item, but no enclusing container
+
+
+IMAGE_PATH = None
+IMAGE_WEB_PATH = None
+
+
+def load_plugins(filename):
+    global Bookmark, BulletedList, BulletedListItem, ChildPageBlock, CodeBlockFenced, \
```
johndgiese, line 34: I feel like there may be a way to do this without needing to repeat all of these names here and in the for loop... that said, this approach is straightforward and completely fine too---perhaps all of the alternatives are more complicated.

n2y/converter.py
```diff
@@ -0,0 +1,466 @@
...
+        else:
+            raise NotImplementedError(f"Unknown plugin type {key}")
+
+
+def load_block(client: Client, id, get_children=True):
+    block = client.get_block(id)
+    return parse_block(client, block, get_children)
+
+
+def parse_block(client: Client, block, get_children=True):
```
johndgiese, line 90: It feels a little weird to me that we're passing the notion client down into all of the blocks. I may be off base with this, but it feels like it would be nice if we could conceptually divide the code into two discrete stages:

1. Grabbing data from Notion
2. Converting this data into our own format

Separating it out into two stages like this would remove the need for patching the client in the tests.

I guess one advantage of intermixing them is that users could write a plugin that decides to skip over certain blocks that it's not interested in, thus saving API calls.

What do you think? Would it be possible to split up the work into two discrete stages, or is there some reason why intermixing the data-fetching with the data conversion is better.

n2y/converter.py
```diff
@@ -0,0 +1,466 @@
...
+    elif block['type'] == "toggle":
+        return Toggle(client, block, get_children)
+    else:
+        # TODO: add remaining block types
+        raise NotImplementedError(f"Unknown block type {block['type']}")
+
+
+class Block():
+    def __init__(self, client: Client, block, get_children=True):
+        # print(f"Creating {block['type']}")
```
johndgiese, line 124: Dead comment?

n2y/main.py
```diff
@@ -1,36 +1,100 @@
...
+    elif args.output == 'yaml':
+        export_yaml(client, database_id, options=args)
+
     return 0
 
 
+def export_markdown(client, database_id, options):
+    raw_rows = client.get_database(database_id)
+
+    if options.name_column not in raw_rows[0]:
```
johndgiese, line 54: I like that you have an argument that lets you select the name_column.

Instead of defaulting to the column with a name of "title", which may not exist, I think you can default to the column with the "title" data type. I believe this is guaranteed to exist and usually is title-like.

What happens now if there are

n2y/main.py
```diff
@@ -1,36 +1,100 @@
...
+        print(f"Database does not contain the column \"{options.name_column}\". "
+              f"Please specify the correct name column using the --name-column flag.")
```
johndgiese, line 56: It would be nice if you list the available column names in the error message.

tests/test_converter.py
```diff
@@ -0,0 +1,841 @@
+from unittest import mock
+
+import pytest
```
johndgiese, line 3: I'm really happy to see all of these tests! nice work!

README.md
```diff
@@ -22,4 +22,52 @@ Copy the link for the database you'd like to export to YAML. Note that linked da
...
 
-By default, `n2y` will simplify the JSON blobs returned by Notion. You can dump the raw data using the `--raw` argument.
+## Converting a Database to markdown files
+
+```
+n2y -o markdown DATABASE_LINK
+```
+
+
+## Plugins
```
johndgiese, line 32: 👍 for updating the readme!

n2y/simplify.py
```diff
@@ -87,6 +105,8 @@ def simplify_people(data):
 
 
 def simplify_date(data):
+    if data is None:
```
johndgiese, line 108: 👍

n2y/simplify.py
```diff
@@ -41,12 +42,29 @@ def simplify_property(prop):
...
+        return prop["files"]
     else:
         # TODO: add remaining column types
         raise NotImplementedError()
 
 
 def simplify_title(data):
+    if data is None or len(data) == 0:
+        return None
     return data[0]["plain_text"]
```
johndgiese, line 70: Probably need to update the `simplify_select` to account for `data` being `None` too. I ran into this when testing the tool on our website.

n2y/main.py
```diff
@@ -1,36 +1,100 @@
...
+            f.write('---\n')
+            f.write(yaml.dump(meta))
+            f.write('---\n\n')
+            f.write(markdown)
+
+
+def export_yaml(client, database_id, options):
+    raw_rows = client.get_database(database_id)
+
+    if not(options.name_column in simplify.flatten_database_row(raw_rows[0])):
```
johndgiese, line 78:
```suggestion
    if not options.name_column in simplify.flatten_database_row(raw_rows[0]):
```

n2y/main.py
```diff
@@ -1,36 +1,100 @@
...
+    if args.output == 'markdown':
+        export_markdown(client, database_id, options=args)
+    elif args.output == 'yaml':
+        export_yaml(client, database_id, options=args)
+
     return 0
 
 
+def export_markdown(client, database_id, options):
+    raw_rows = client.get_database(database_id)
```
johndgiese, line 52: It seems like there's bit of duplicated code at the top of `export_markdown` and `export_yaml` that could be avoided by pulling this code up a level.

johndgiese (commented):
Really nice work Ilya, the code is looking great! I played around with the Commandline tool a little bit. In addition to addressing my comments, if you could try pulling down this database:

https://www.notion.so/innolitics/01dee73fef7e4dd2a02a2a50b6848ba1?v=77043ea15048428ca095601714176e0f

That would be great. The access token is in 1password under the "Notion "Website" Integration Token" entry. The integration is already added to the page and has read-only access.

n2y/notion.py
```diff
@@ -40,6 +40,78 @@ def _get_database(self, url, database_id):
             raise ValueError(f"Unable to find database with id '{database_id}'")
         return response.json()
 
+    # recursively get block children
+    def get_block_children(self, block_id, recursive=False):
```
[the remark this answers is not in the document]
xaker00: You can use this to get the full JSON without running the converter. If the converter is used, then it gets blocks one level at a time.

n2y/simplify.py
xaker00: Done

n2y/main.py
```diff
@@ -1,36 +1,100 @@
...
+
+    if options.name_column not in raw_rows[0]:
+        print(f"Database does not contain the column \"{options.name_column}\". "
+              f"Please specify the correct name column using the --name-column flag.")
+        exit(1)
+
+    for row in raw_rows:
+        meta = simplify.flatten_database_row(row)
+        print(f"Processing {meta[options.name_column]}", file=sys.stderr)
```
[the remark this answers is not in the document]
xaker00: Since YAML is being sent to stdout, do not want to pollute that with progress and warning messages.

xaker00 (commented):
Ready for round 2

n2y/converter.py
johndgiese: That said, it wouldn't surprise me either if any solution we come up with has other downsides and the straightforward approach you have here is verbose but very simple. Thus, unless you're just curious now about finding another way, I'm fine just leaving as is too.

n2y/converter.py
johndgiese: That makes sense! If you haven't already, I think it would be worth copying your response to my GitHub comment into a code comment somewhere. It seems like the sort of high-level "why" question that's unlikely to go stale and may help future people working on the project understand the architecture.

n2y/converter.py
```diff
@@ -0,0 +1,466 @@
...
+    def to_pandoc(self):
+        return BulletList([i.to_pandoc() for i in self.items])
+
+
+class ToDoItem(BulletedListItem):
+    def __init__(self, client: Client, block, get_children=True):
+        self.type = 'to_do_list'
+        super().__init__(client, block, get_children)
+        if self.checked:
+            self.text.text[0].plain_text.text = '☒ ' + self.text.text[0].plain_text.text
```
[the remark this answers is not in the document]
johndgiese: Ahh, I see! That makes sense! I'd add a comment about this in the code.

n2y/converter.py
```diff
@@ -0,0 +1,466 @@
...
+        if obj['type'] == "file":
+            self.type = "file"
+            self.url = obj['file']['url']
+            self.expiry_time = obj['file']['expiry_time']
+        elif obj['type'] == "external":
+            self.type = "external"
+            self.url = obj['external']['url']
+
+    def download(self):
+        # TODO: append created time as hex to end of file to prevent collisions?
```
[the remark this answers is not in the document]
xaker00: Sorry, I was misreading my own code. For some reason, I thought this was naming the page and not file attachments. md5 should be fine for files since they changed less frequently.

n2y/converter.py
xaker00: I think using `globals()` with `issubclass()` would be a better solution than the current one. Using `issubclass( Block)` would prevent the plugins from overriding things they should not.

n2y/main.py
xaker00: Duplicate code consolidated.

xaker00:
Updated per comments.

n2y/converter.py
```diff
@@ -31,61 +31,32 @@
...
-            TableBlock = value
-        elif key == "ToDo":
-            ToDo = value
-        elif key == "ToDoItem":
-            ToDoItem = value
-        elif key == "Toggle":
-            Toggle = value
+        if key in globals():
+            # plugins can only override classes in this file that are derrived from a Block
+            if globals()[key].__name__ in [b.__name__ for b in value.__bases__] \
```
johndgiese, line 41: I think this line can be simplified by using some intermediate variables (e.g., for `globals()[key]` and `[b.__name__ for b in value.__bases__]`.

n2y/main.py
```diff
@@ -1,36 +1,124 @@
...
+
     client = notion.Client(ACCESS_TOKEN)
+
     raw_rows = client.get_database(database_id)
-    if args.raw:
-        print(yaml.dump(raw_rows))
-        return 0
+    first_row_flattened = simplify.flatten_database_row(raw_rows[0])
+
+    def available_columns():
```
johndgiese, line 46: Why use a function for this instead of just calculating the value? Is this to avoid calculating the value if it doesn't end up being used?

johndgiese (approved):
I had a few small suggestions, but feel free to merge once they're addressed!

n2y/main.py
xaker00: Yes, there are two places where the available columns may be shown, but there is usually no need to generate this list. Also, I imagine this function will grow in future versions as we would want deeper checks.

PR #9, opened 2022-03-14
Express

sHermanGriffiths:
Reconfigured code to utilize the `Math`, `InlineMath`, & `DisplayMath`, types from pandoc when converting Notion to Markdown. Also added the `tex_math_dollars` extension to the format given to `pandoc.write()` to convert equations with one "\" instead of two


johndgiese (commented):
@sHermanGriffiths please be sure the documentation is updated as appropriate and there are no dead files that a re included. E.g., I don't think `dict/n2y-0.1.0-py3.9.egg` should be included. Otherwise, please request a review from @xaker00 to check the code and tests.

xaker00 (approved):
Looks good!

PR #10, opened 2022-03-17
Add target directory arguement

sHermanGriffiths:
Add `--target` or `-t` argument to n2y so that users can directly specify target directory for markdown files.


build/lib/n2y/converter.py
```diff
@@ -0,0 +1,473 @@
+import importlib.util
```
johndgiese, line 1: @sHermanGriffiths I don't think the `build` directory should be included in this PR. Please review all of your files before committing them. Also, I recommend against using the `-a` command with `git commit` since it makes it too easy to accidentally add files like this. Also, feel free to update the `.gitignore` file as needed.

n2y/main.py
```diff
@@ -95,11 +97,19 @@ def export_markdown(client, raw_rows, options):
...
                 markdown = pandoc.write(pandoc_output, format='gfm+tex_math_dollars') \
                     .replace('\r\n', '\n')  # Deal with Windows line endings
-                # reformat equations to have one backslash instead of 2
+
+                # normalize target path and create it if it doesn't exist
+                target = options.target
+                if not target[(len(target)-1)] == '/':
+                    target += '/'
+                    options.target = target
+                if not os.path.exists(target):
```
johndgiese, line 106: I think it's cleaner (and avoids a small race condition which can occur if the directory is created in between checking if it exists and creating it) if you use `exists_ok` instead of checking if the path exists and then creating it.

n2y/main.py
```diff
@@ -95,11 +97,19 @@ def export_markdown(client, raw_rows, options):
             if pandoc_output:
                 markdown = pandoc.write(pandoc_output, format='gfm+tex_math_dollars') \
                     .replace('\r\n', '\n')  # Deal with Windows line endings
-                # reformat equations to have one backslash instead of 2
+
+                # normalize target path and create it if it doesn't exist
+                target = options.target
+                if not target[(len(target)-1)] == '/':
+                    target += '/'
```
johndgiese, line 104: Do you need to add the `/` to the end of the path? I don't think `os.path.makedirs` should care if there is a path or not.

n2y/main.py
```diff
@@ -95,11 +97,19 @@ def export_markdown(client, raw_rows, options):
...
+                    options.target = target
+                if not os.path.exists(target):
+                    os.makedirs(target)
+                    options.target = target
 
                 # sanitize file name just a bit
                 # maybe use python-slugify in the future?
-                with open(f"{filename}.md", 'w') as f:
+                with open(f"{options.target}{filename}.md", 'w') as f:
```
johndgiese, line 112: I would use `os.path.join` here instead of combining the filename with the path like this. Not only is it cleaner but it also will work on different operating systems. See https://docs.python.org/3/library/os.path.html#os.path.join

johndgiese (changes requested):
Nice work! See comments!

n2y/main.py
```diff
@@ -21,6 +21,8 @@ def main():
     parser.add_argument("--image-path", help="Specify path where to save images")
     parser.add_argument("--image-web-path", help="Web path for images")
     parser.add_argument("--plugins", help="Plugin file")
+    parser.add_argument("--target", '-t', default='./',
```
johndgiese, line 24: You may want to look at some other command line tools to see what other people use to indicate the output directory's location. When possible, it's nice to make the command line arguments line up with what other tools do. You may have done this already.

n2y/main.py
xaker00: +1 for `os.path.join` There should not be any hardcoded path separators.

.gitignore
```diff
@@ -2,3 +2,4 @@
 __pycache__
 n2y.egg-info/
 .DS_Store
+build/
```
xaker00, line 5: Good idea

xaker00 (changes requested):
Should images be saved in the same folder as the markdown if an image folder is not explicitly specified? I can imagine a user being surprised to get images in the working directory when they specify a target location. See https://github.com/innolitics/n2y/blob/ed0c40f2540e767b64ad509115b0c5a738de3f8e/n2y/converter.py#L411

n2y/main.py
sHermanGriffiths: it was moreso for the format string (`f"{options.target}{filename}.md"`) on line 112 but now that it uses `os.path.join` it's unnecessary.

.gitignore
sHermanGriffiths: Just made an addition at lines 38-39, is that what you meant?

sHermanGriffiths:
> Should images be saved in the same folder as the markdown if an image folder is not explicitly specified? I can imagine a user being surprised to get images in the working directory when they specify a target location. See
> 
> https://github.com/innolitics/n2y/blob/ed0c40f2540e767b64ad509115b0c5a738de3f8e/n2y/converter.py#L411

Just made an addition at lines 38-39, is that what you meant?

n2y/main.py
johndgiese: Although modern versions of windows _should_ be fine with the `/` instead of `\` for a directory separator, I'd still suggest changing the default from `./` to just `.` since that seems cleaner and devs may think that `./` is invalid.

n2y/main.py
```diff
@@ -95,11 +99,14 @@ def export_markdown(client, raw_rows, options):
             if pandoc_output:
                 markdown = pandoc.write(pandoc_output, format='gfm+tex_math_dollars') \
                     .replace('\r\n', '\n')  # Deal with Windows line endings
-                # reformat equations to have one backslash instead of 2
+
+                # create target path if it doesn't exist
```
johndgiese, line 103: I don't think this comment is necessary since it's pretty clear from `os.makedirs(options.target, exist_ok=True)` what the line does

johndgiese (approved):
Feel free to merge once my two comments have been addressed.

n2y/main.py
```diff
@@ -33,6 +35,8 @@ def main():
 
     database_id = notion.id_from_share_link(args.database)
 
+    if not args.image_path:
+        args.image_path = args.target
```
xaker00, line 39: This works, but it's probably better not to change the `args` object. A better way:

```python
if args.image_path:
    converter.IMAGE_PATH = args.image_path
else:
    converter.IMAGE_PATH = args.target
```

Feel free to change it or keep it as-is. I'm probably getting a little nit-picky.

xaker00 (approved):

n2y/main.py
johndgiese: Great point @xaker00. Yeah, @sHermanGriffiths most of the time it's best to avoid mutating a complex object that was generated upstream from your code.

PR #11, opened 2022-03-24
Dupefile warning

sHermanGriffiths:
Prevents collisions in markdown and image file paths


n2y/converter.py
```diff
@@ -384,46 +385,73 @@ def to_pandoc(self):
 
 
 class ImageBlock(Block):
+    # `ImageBlock` Class Object Initialization
```
johndgiese, line 388: Why do you think this comment is worth including?

johndgiese: Actually, I'd go through and ask this question about all of the added comments. In general, comments should focus on why the code does what it does and should generally avoid explaining what the code does, since good naming can already do that.

n2y/converter.py
```diff
@@ -384,46 +385,73 @@ def to_pandoc(self):
...
+        # Set Values For The `type` And `url` Keys Based On That Of `obj` Param
+        for type in ["file", "external"]:
+            if obj['type'] == type:
+                self.type = type
+                self.url = obj[type]['url']
+                # Only "file" Typed `obj`s Have `expiry_time`, So Set That Key If It's There
+                self.expiry_time = obj[type]['expiry_time'] if type == "file" else None
+
+    # Pevents Collisions In The File Paths N2Y Creates
+    def create_unique_url(self, url):
```
johndgiese, line 420: Does this function create unique URLs or unique file names? If the former is the case, I don't see how/why we would want to create unique URLs.

johndgiese:
@sHermanGriffiths what was the motivation for this PR? Is this solving a problem you ran into when working on this issue?

https://github.com/innolitics/innolitics.github.io/issues/328

If not, I would put this PR on pause and resume work on that issue for now.

sHermanGriffiths:
Yes, It's the version of n2y with the changes necessary for https://github.com/innolitics/innolitics.github.io/issues/328 to work as expected. It also inadvertently handles a previously raised issue in this repository.

n2y/converter.py
sHermanGriffiths: With that objective in mind I just deleted them all.

n2y/converter.py
sHermanGriffiths: I changed the name

n2y/converter.py
```diff
@@ -389,41 +390,53 @@ def __init__(self, client: Client, block, get_children=True):
...
-            self.url = obj['file']['url']
-            self.expiry_time = obj['file']['expiry_time']
-        elif obj['type'] == "external":
-            self.type = "external"
-            self.url = obj['external']['url']
+        for type in ["file", "external"]:
+            if obj['type'] == type:
+                self.type = type
+                self.url = obj[type]['url']
+                self.expiry_time = obj[type]['expiry_time'] if type == "file" else None
```
xaker00, line 411: I'm not sure that you need the for loop and the if statement. This might me a better way to do the same thing.

```python
self.type = obj['type']
self.url = obj[self.type]['url']
self.expiry_time = obj[self.type]['expiry_time'] if self.type == "file" else None
```

n2y/converter.py
```diff
@@ -29,6 +29,7 @@
 
 IMAGE_PATH = None
 IMAGE_WEB_PATH = None
+NAMING_SYSTEM = {}
```
xaker00, line 32: Is this used only in this module? If so you should follow the naming convention for private variables. See https://peps.python.org/pep-0008/#descriptive-naming-styles

xaker00 (changes requested):
Good effort. The problem is that I think the approach is flawed. This code simply increments when it finds another occurrence. That would work if the blocks always came in the same order. This is not the case. Pages may come in a different order and images may be moved around on a page. There needs to be a more robust system. I don't think that we want to change our page links every time the order changes.

You might want to simply append the notion block id to all files. They should be unique, and you can use them to uniquely trace to the original block if you ever wanted to. 

Another approach would be to take a md5 sum of the contents and use that in the filename. (suggested by David earlier)

johndgiese:
@xaker00 I agree with your assessment. In particular, we would not want the filenames to change based on the order that the images were seen.

@sHermanGriffiths I realized I'd asked you to look into this feature, but let's pause working on it for now. Since we are primarily interested in n2y for our own use case and since we can edit the filenames in Notion, I don't think name collisions with images will be a problem for a while. I'm not sure what I was thinking when I asked you to work on this, so I apologize about that.

PR #14, opened 2022-03-29
Don't use the name column argument when generating YAML

johndgiese:
@xaker00 it didn't seem like we need to use the name column argument when we generate YAML. What do you think?


xaker00:
I agree. It's probably not needed. The only useful thing that option did was make it the first field in YAML just in case a human wanted to look at it. (it might still be the first field for other reasons) Keeping it might cause more problems.

xaker00 (approved):

PR #18, opened 2022-03-30
Correct Annotations

sHermanGriffiths:
Spaces at the end and beginning of certain annotations would cause a glitch in the conversion to markdown. The code added in this branch is meant to correct those glitches

### DIFF
--- a/n2y/converter.py
+++ b/n2y/converter.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import importlib.util
 from os import path, makedirs
 from shutil import copyfileobj
@@ -162,19 +163,49 @@ class Annotations():
     def __init__(self, block):
         for key, value in block.items():
             self.__dict__[key] = value
+    
+    def strip(self, target, annotations):
+        prependages = []
+        appendages = []
+        print(f"TARGET 1: {target}")
+        blank_space = [Space(), SoftBreak()]
+        while target[0] in blank_space:
+            prependages.append(target.pop(0))
+        while target[-1] in blank_space:
+            appendages.append(target.pop(-1))
+        result = []
+        print(f"TARGET 2: {target}")
+        if prependages:
+            for i in range(len(prependages) - 1, -1, -1):
+                result.append(prependages[i])
+        nested_annotations = False
+        if self.code:
+            target = Code(("", [], []), target)
+            nested_annotations = True
+        if self.bold:
+            target = Strong([target]) if nested_annotations else Strong(target)
+            nested_annotations = True
+        if self.italic:
+            target = Emph([target]) if nested_annotations else Emph(target)
+            nested_annotations = True
+        if self.underline:
+            target = Underline([target]) if nested_annotations else Underline(target)
+            nested_annotations = True
+        if self.strikethrough:
+            target = Strikeout([target]) if nested_annotations else Strikeout(target)
+            nested_annotations = True
+        result.append(target)
+        if appendages:
+            for i in range(len(appendages) - 1, -1, -1):
+                result.append(appendages[i])
+        print(f"RESULT: {result}")
+        return result
 
     def apply_pandoc(self, target):
         result = target
-        if self.code:
-            result = [Code(("", [], []), result)]
-        if self.bold:
-            result = [Strong(result)]
-        if self.italic:
-            result = [Emph(result)]
-        if self.underline:
-            result = [Underline(result)]
-        if self.strikethrough:
-            result = [Strikeout(result)]
+        annotations = [self.code, self.bold, self.italic, self.underline, self.strikethrough]
+        if True in annotations:
+            result = self.strip(result, annotations)
         return result
 
 
--- a/n2y/converter.py
+++ b/n2y/converter.py
@@ -237,8 +237,8 @@ class RichText():
                 return self.annotations.apply_pandoc(self.plain_text.text)
             elif 'href' in self.__dict__ and self.href:
                 # links
-                return [
-                    Link(('', [], []),
+                return [Link(
+                    ('', [], []),
                     self.annotations.apply_pandoc(self.plain_text.to_pandoc()),
                     (self.href, ''))]
             else:
@@ -473,19 +473,23 @@ class TableBlock(Block):
         # Notion does not have cell alignment or width options, sticking with defaults.
         colspec = [(AlignDefault(), ColWidthDefault()) for _ in range(self.table_width)]
         table = Table(
-            ('', [], []), # attr
-            Caption(None, []), # caption
+            ('', [], []),  # attr
+            Caption(None, []),  # caption
             colspec,
-            TableHead(('', [], []), header_rows), # table header
-            [TableBody(('', [], []), RowHeadColumns(row_header_columns), [], children)], # table body
-            TableFoot(('', [], []), [])) # table footer
+            TableHead(('', [], []), header_rows),  # table header
+            [TableBody(
+                ('', [], []),
+                RowHeadColumns(row_header_columns),
+                [],
+                children)],  # table body
+            TableFoot(('', [], []), []))  # table footer
         return table
 
 
 class RowBlock(Block):
     def to_pandoc(self):
-        cells = [
-            Cell(('', [], []),
+        cells = [Cell(
+            ('', [], []),
             AlignDefault(),
             RowSpan(1),
             ColSpan(1),
--- a/n2y/converter.py
+++ b/n2y/converter.py
@@ -166,10 +166,8 @@ class Annotations():
 
     def handler(self, annotation, target):
         handler = {
-            'code': [Code(("", [], []), [target]), Code(("", [], []), target)],
             'bold': [Strong([target]), Strong(target)],
             'italic': [Emph([target]), Emph(target)],
-            'underline': [Underline([target]), Emph(target)],
             'strikethrough': [Strikeout([target]), Strikeout(target)]
         }
         return handler[annotation]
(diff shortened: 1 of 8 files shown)


tests/test_converter.py
```diff
@@ -454,6 +458,140 @@ def test_strikeout_word():
...
+                        'italic': False,
+                        'strikethrough': False,
+                        'underline': False,
+                        'code': False,
+                        'color': 'default'},
+                    'plain_text': '?',
+                    'href': None}]}}
+    obj = converter.ParagraphBlock(None, input, get_children=False)
+    pandoc_output = obj.to_pandoc()
+    print(pandoc_output)
```
johndgiese, line 579: Leftover print statement?

tests/test_converter.py
```diff
@@ -415,9 +415,13 @@ def test_bold_italic_word():
 
     obj = converter.ParagraphBlock(None, input, get_children=False)
     pandoc_output = obj.to_pandoc()
-
-    assert pandoc_output == Para([Str("A"), Space(), Emph([Strong([Str("bold-italic")])]),
-                                 Space(), Str("word.")])
+    print(f"PANDOC_OUTPUT: {pandoc_output}")
```
johndgiese, line 418: Leftover print statement?

n2y/converter.py
```diff
@@ -163,18 +164,48 @@ def __init__(self, block):
...
 
+    def handler(self, annotation, target):
+        handler = {
+            'bold': [Strong([target]), Strong(target)],
+            'italic': [Emph([target]), Emph(target)],
+            'strikethrough': [Strikeout([target]), Strikeout(target)]
+        }
+        return handler[annotation]
+
+    def strip(self, target):
```
johndgiese, line 175: I'll leave it to @xaker00 to review this algorithm in detail, but it looks nice at first glance!

tests/test_converter.py
```diff
@@ -454,6 +458,140 @@ def test_strikeout_word():
     assert newline_lf(markdown_output) == expected_markdown
 
 
+def test_annotated_spaces():
+    input = {
+        'object': 'block',
```
johndgiese, line 463: Is it necessary to include all of this meta data, e.g., `created_time`, in the test fixture here? It seems unrelated to what we're testing. If it is needed, perhaps there's some way we can create a mock utility so that we can hide this complexity outside of the tests. If other tests include this metadata unnecessarily, I suggest we refactor those other tests as well.

sHermanGriffiths: no, I can certainly and will delete some of it

tests/test_converter.py
sHermanGriffiths: yea, there were a bunch haha

tests/test_converter.py
```diff
@@ -454,6 +458,140 @@ def test_strikeout_word():
...
+            'id': 'ad6d20a2-36e9-4b6d-a49a-7f71f8ba9af6'},
+        'last_edited_by': {
+            'object': 'user',
+            'id': 'ad6d20a2-36e9-4b6d-a49a-7f71f8ba9af6'},
+        'has_children': False,
+        'archived': False,
+        'type': 'paragraph',
+        'paragraph': {
+            'color': 'default',
+            'text': [
```
johndgiese, line 468: This is a big object to read through to understand what the test is doing. What do you think about this idea:

Write a utility function that can generate the objects for us using an easier to read format. I've used this technique often in other tests:

```python
input = annotated_text([("this ", "b"), ("is ", "bi"), ("a", "I"), (" test", "s")])
```

I think you get the idea.

Also, although I didn't read through the test data carefully, I suspect you're testing a few different things all in one big test. I suspect you did this in part because the current test data fixture is so big it felt bad to copy it a few times for multiple tests. Once you've created this little function I showed you, you can split apart this big test into a few smaller tests that all have good function names. Each test should test one particular situation. You may need several of them to cover all the corner cases.

Let me know if any of this is unclear and I'd be happy to help.

johndgiese: I'd also update the other tests to use this new tool, where appropriate.

sHermanGriffiths: makes sense, I'll comment when it's done

johndgiese: Very close, nice work! everything you've added looks great but please reread my comment carefully and be sure you addressed everything I brought up. Also, I think you can use tuple unpacking to in the for loop so that you can give block[0] and block[1] better names.

tests/test_converter.py
```diff
@@ -6,18 +6,46 @@
...
+def generate_annotated_obj(arr):
+    obj = {
+        'object': 'block',
+        'has_children': False,
+        'archived': False,
+        'type': 'paragraph',
+        'paragraph': {
+            'color': 'default',
+            'text': []}}
+    for block in arr:
```
johndgiese, line 31: You can expand the tuple in the for loop directly!

Also, `arr` is not a very good name. I'd change it to something else, like `text_blocks_descriptors`:

```suggestion
    for (text, annotations) in text_blocks_descriptors:
```

johndgiese (approved):
@sHermanGriffiths I had one small suggested changed which I'd ask you to make, but otherwise this looks great! I'll defer to @xaker00 if he wants to review the algorithm itself.

n2y/converter.py
```diff
@@ -163,18 +164,48 @@ def __init__(self, block):
...
+        prependages = deque()
+        appendages = deque()
+        result = []
+        blank_space = [Space(), SoftBreak()]
+        nested_annotations = False
+        while target[0] in blank_space:
+            prependages.appendleft(target.pop(0))
```
xaker00, line 182: Logic error here. `prependages` ends up reversed here

n2y/converter.py
xaker00, line 175: `strip` does more than strip. It does the conversion also.

xaker00 (changes requested):
Good job improving and cleaning up tests. I would suggest a simpler version of the `Annotations` class:

```python
class Annotations():
    def __init__(self, block):
        for key, value in block.items():
            self.__dict__[key] = value

    def apply_pandoc(self, target):
        prependages = deque()
        appendages = deque()

        blank_space = [Space(), SoftBreak()]
        while target[0] in blank_space:
            prependages.append(target.pop(0))
        while target[-1] in blank_space:
            appendages.appendleft(target.pop(-1))

        def listify(obj):
            return obj if type(obj) == list else [obj]

        result = target
        if self.code:
            result = listify(Code(("", [], []), result))
        if self.bold:
            result = listify(Strong(result))
        if self.italic:
            result = listify(Emph(result))
        if self.underline:
            result = listify(Underline(result))
        if self.strikethrough:
            result = listify(Strikeout(result))

        return list(prependages) + result + list(appendages)
```

n2y/converter.py
sHermanGriffiths: it's actually on purpose. I wanted to use the deque like a stack. I figure this way it left appends and now the first thing served in the for loop is the last thing added.

sHermanGriffiths: Thanks, I was visualizing it incorrectly. The appendages _should_ be appended left though, right?

n2y/converter.py
```diff
@@ -163,18 +164,48 @@ def __init__(self, block):
...
+                result.append(i)
+        result.append(target)
+        if appendages:
+            for i in appendages:
+                result.append(i)
+        return result
+
     def apply_pandoc(self, target):
         result = target
+        annotations = [self.bold, self.italic, self.strikethrough]
```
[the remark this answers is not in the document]
sHermanGriffiths: Yes but only the bold, italic, and strikethrough annotations can't affect leading and trailing spaces in gfm so I figure only change those in order to make our conversion as close to the original document as possible, only making changes when absolutely neccessary.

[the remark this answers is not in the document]
johndgiese: I like the idea of making the minimal number of changes to the original document as possible. We may not even need to trim the space for strikethrough _if_ struck out spaces are visible in Notion. I think we definitely need to handle bold and italic because bolded spaces look just like unbolded spaces and italicized spaces look just like unitalicized spaces.

n2y/converter.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
johndgiese, line 1: What is this import for?

n2y/converter.py
xaker00: You have two loops one going forward, and the other backward. When you reconstruct the result in the end, all of the pieces need to be arranged left to right. This is why on the reverse loop the items need to be added to the other side. In effect reversing a reversed list and making it correct.

n2y/converter.py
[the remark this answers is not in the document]
xaker00: leading and trailing spaces confuse markdown renderers. In the case of italic, a leading space will add a bullet instead of making the word italic.

This is how vscode renders:
image

n2y/converter.py
xaker00: That is vscode trying to be helpful. Have to watch out for these. Please delete @sHermanGriffiths.

In vscode if you try to use a variable without first assigning it, vscode will try to find something that it can import to match that name. This is both helpful and annoying. Vscode is awesome for some things, but not as smart as PyCharm when it comes to python.

xaker00 (approved):
Looks good. Can be merged after removing the extra import mentioned here: https://github.com/innolitics/n2y/pull/18#discussion_r841739491.

n2y/converter.py
sHermanGriffiths: deleted

n2y/converter.py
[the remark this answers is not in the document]
sHermanGriffiths: > We may not even need to trim the space for strikethrough _if_ struck out spaces are visible in Notion.

@johndgiese I know that Notion can render struck out spaces just fine but the final markdown conversion renders as:
 ~~ STRIKETHROUGH ~~
Instead of something that looks like:
-~~STRIKETHROUGH~~-

With that in mind, I think we should still correct them. What do you think?

johndgiese:
Agreed!

Issue #13, opened 2022-03-25
Improve handling of articles with blank slug value


johndgiese:
N2y already will ignore rows in a notion database who have a blank value in the specified "slug column".

It does appear, however, that at least one file named `.md` is created when there are blank slug values. Perhaps this is caused if the slug column just has whitespace?

Also, we should document this behavior here https://github.com/innolitics/n2y/blob/main/n2y/main.py#L26-L28

Finally, we should either remove the warning associated with blank files (since this is now expected behavior and there may be many rows with blank slugs) OR we should condense the warnings into a single line of text that says something like "Ignored one or more rows with blank slug column values."

PR #19, opened 2022-04-05
Errlog

sHermanGriffiths:
Replace print statements by raising errors and making use of Python's `logging` library. Also, print fewer warnings to standard out. Instead, create an amalgamation of currently intermediate warnings and log it at the end of the function.


n2y/converter.py
```diff
@@ -1,18 +1,18 @@
-from collections import deque
+import re
+import logging
+import requests
```
johndgiese, line 3: See the PEP8 section on module import ordering; I think `requests`, since it's a third party library, should go in the second batch of imports.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
-              file=sys.stderr)
-        print("Available column(s): ",
-              # only show columns that have strings as possible options
-              ", ".join(available_columns()), file=sys.stderr)
-        return False
+        raise NotImplementedError("%s%s%s" % (
+            f"ERROR: Database Does Not Contain The Column \"{name_column}\".\n",
+            "Please Specify The Correct Name Column Using The --name-column Flag.\n",
+            # only show columns that have strings as possible options
+            "Available Column(s): " + ", ".join(available_columns())))
```
xaker00, line 78: Raising an exception displays the stack trace and looks like the program crashed. I don't think the program should "crash". Previous behavior was good. Display the problem and possible solution then exit with a non-zero return value.

Also, if we wanted to throw an exception, `NotImplementedError` is not the correct one here. See https://docs.python.org/3/library/exceptions.html for exception types.

n2y/converter.py
```diff
@@ -46,9 +47,11 @@ def load_plugins(filename):
                     and issubclass(class_to_replace, Block):
                 globals()[key] = value
             else:
-                print(f"Cannot import plugin \"{key}\" since it not derrived from a known class.")
+                logger.warning("%s %s" % (
```
johndgiese, line 50: I don't think you should need to use the `%` operator here. Instead, you can use `logger.warning("%s %s",  f"Warning: Cannot Import Plugin \"{key}\"", "Because It Is Not Derrived From A Known Class.")`

Or, even better, you can use:

`logger.warning('Warning: Cannot import plugin "%s" because it is not derived from a known class.', key)`

This is better for a few reasons:

1. There's no need to use %s formatting for string literals; you can just combine the strings directly.
2. It avoids using f-strings in the log messages; this is good because it avoids wasting processing time formatting the string if the logging settings are set such that the message won't be formatted. By using the %s syntax for the key, the logging module can be smart enough to skip bothering with this if the message would just get thrown out anyway.
3. It doesn't capitalize all of the words in the message (usually you would write messages as if they're a sentence in regular English.

n2y/converter.py
```diff
@@ -46,9 +47,11 @@ def load_plugins(filename):
...
+                    f"Warning: Cannot Import Plugin \"{key}\"",
+                    "Because It Is Not Derrived From A Known Class."))
         else:
-            raise NotImplementedError(f"Unknown plugin type {key}.")
+            raise NotImplementedError(f"Unknown Plugin Type \"{key}\".")
```
johndgiese, line 54: Shouldn't be capitalized here

n2y/main.py
```diff
@@ -1,37 +1,41 @@
 import os
-import sys
-import argparse
 import re
-
+import sys
 import yaml
 import pandoc
-
+import logging
```
johndgiese, line 6: The import ordering is no longer PEP8 compliant

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
     # make sure the title column exists
     if name_column not in first_row_flattened:
-        print(f"Database does not contain the column \"{name_column}\". "
-              f"Please specify the correct name column using the --name-column flag.",
-              file=sys.stderr)
-        print("Available column(s): ",
-              # only show columns that have strings as possible options
-              ", ".join(available_columns()), file=sys.stderr)
-        return False
+        raise NotImplementedError("%s%s%s" % (
```
johndgiese, line 74: I wouldn't recommend using the `%` operator like this. The implicit string concatenation used in the original version is cleaner IMO.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
+
+                pandoc_output = converter.load_block(client, row['id']).to_pandoc()
+                # do not create markdown pages if there is no page in Notion
+                if pandoc_output:
+                    logger.warning(f'Processing Page "{meta[options.name_column]}"')
+                    markdown = pandoc.write(pandoc_output, format='gfm+tex_math_dollars') \
+                        .replace('\r\n', '\n')  # Deal with Windows line endings
+
+                    # create target path if it doesn't exist
+                    # if not os.path.exists(options.target):
```
johndgiese, line 112: Dead code? also I think the comment doesn't say anything that isn't obvious in the code already.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
+    types_skipped = 0
+    prefixes = ("", " & ", ", & ")
+    for key in skips.keys():
+        count = skips[key]
+        if count > 0:
+            msg = msg if key != "Duplicate" and types_skipped < 2 else msg.replace(" &", ",")
+            msg += f"{prefixes[types_skipped]}{count} {key}"
+            types_skipped += 1
+
+    msg == "" or logger.warning(f"WARNING: {msg} Page(s) Skipped")
```
johndgiese, line 144: I'd rely on the logger message configuration to indicate that this particular message is a warning, rather than including the string `WARNING: ` here. This allows the formatting to be handled in a centralized way and will ensure the messages are consistent.

n2y/simplify.py
```diff
@@ -1,8 +1,10 @@
 """
 Simplify data that has been returned from the Notion API.
 """
+import logging
 import re
-import sys
+
+logger = logging.getLogger('n2y.simplify')
```
johndgiese, line 7: Usually people will just use `logger = logging.getLogger(__name__)`, since this will get the module name automatically.

n2y/simplify.py
```diff
@@ -45,17 +47,17 @@ def simplify_property(prop):
     elif prop["type"] == "multi_select":
         return [option["name"] for option in prop["multi_select"]]
     elif prop["type"] == "relation":
-        print("WARNING: field type 'relation' is not fully implemented", file=sys.stderr)
+        logger.warning("WARNING: Field Type 'Relation' Is Not Fully Implemented")
```
johndgiese, line 50: Ditto my other comment

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
+            f"ERROR: Database Does Not Contain The Column \"{name_column}\".\n",
+            "Please Specify The Correct Name Column Using The --name-column Flag.\n",
+            # only show columns that have strings as possible options
+            "Available Column(s): " + ", ".join(available_columns())))
 
     # make sure title column is not empty (only the first row is checked)
     if first_row_flattened[name_column] is None:
-        print(f"Column \"{name_column}\" cannot be empty.", file=sys.stderr)
-        return False
+        raise NotImplementedError(f"ERROR: Column \"{name_column}\" Cannot Be Empty.")
```
xaker00, line 82: same as above

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
+                        f.write(yaml.dump(meta))
+                        f.write('---\n\n')
+                        f.write(markdown)
+                else:
+                    logger.debug(f"Skipping Empty Page: {page_name}")
+                    skips['Empty'] += 1
             else:
-                print(f"WARNING: skipping empty page: {page_name}", file=sys.stderr)
+                logger.debug(
+                    "%s %s",
```
johndgiese, line 127: Ditto my other comment.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
     if not isinstance(first_row_flattened[name_column], str):
-        print(f"Column \"{name_column}\" does not contain a string.", file=sys.stderr)
-        print("Available column(s): ",
-              # only show columns that have strings as possible options
-              ", ".join(available_columns()), file=sys.stderr)
-        return False
+        raise NotImplementedError("%s%s" % (
+            f"ERROR: Column \"{name_column}\" Does Not Contain A String.\n",
+            # only show columns that have strings as possible options
+            "Available Column(s): " + ", ".join(available_columns())))
```
xaker00, line 89: same as above

johndgiese (approved):
Nice stab at this @sHermanGriffiths ! I had a lot of little comments for you; they're mostly small things, but it's worth getting the small things right.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
-                    f.write('---\n\n')
-                    f.write(markdown)
+            filename = re.sub(r"[\s/,\\]", '_', page_name.lower())
+            if filename not in file_names:
+                file_names.append(filename)
+
+                pandoc_output = converter.load_block(client, row['id']).to_pandoc()
+                # do not create markdown pages if there is no page in Notion
+                if pandoc_output:
+                    logger.warning(f'Processing Page "{meta[options.name_column]}"')
```
xaker00, line 107:
```suggestion
                    logger.info('Processing Page "%s"', meta[options.name_column])
```

This is not a `warning`. Should be `info` or `debug`. Also, the logger already has a way of composing strings. See https://docs.python.org/3/howto/logging.html#loggers

n2y/main.py
xaker00, line 112: dead code?

n2y/main.py
xaker00, line 144: The logger may be already doing this. Double-check that this is necessary.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
+                        f.write('---\n')
+                        f.write(yaml.dump(meta))
+                        f.write('---\n\n')
+                        f.write(markdown)
+                else:
+                    logger.debug(f"Skipping Empty Page: {page_name}")
+                    skips['Empty'] += 1
             else:
-                print(f"WARNING: skipping empty page: {page_name}", file=sys.stderr)
+                logger.debug(
```
xaker00, line 126: This should be `error` or `warning`.

n2y/main.py
```diff
@@ -61,67 +65,83 @@ def name_column_valid(raw_rows, name_column):
...
+                    "%s %s",
+                    f'Skipping Duplicate Page: "{page_name}",',
+                    'Name Has Already Been Used. Please Rename')
+                skips['Duplicate'] += 1
         else:
-            print("WARNING: skipping page with no name", file=sys.stderr)
+            logger.debug("Skipping Page With No Name")
```
xaker00, line 132: maybe should be `warning`?

n2y/main.py
```diff
@@ -131,7 +151,7 @@ def export_yaml(client, raw_rows):
         markdown = pandoc.write(pandoc_output, format='gfm') if pandoc_output else None
         result.append({**simplify.flatten_database_row(row), 'content': markdown})
 
-    print(yaml.dump(result, sort_keys=False))
+    logger.info(yaml.dump(result, sort_keys=False))
```
xaker00, line 154: Should be a print statement not a logger message. Logging is a side channel. YAML should go to stdout.

xaker00 (commented):
Good progress. Needs more work. See https://docs.python.org/3/library/logging.html#logging.basicConfig for output and formatting of messages. I think we want to send logging messages to stderr.

Feel free to reach out for help.

n2y/main.py
sHermanGriffiths: I need it to be a level that less than the logger's level unless you want the "Processing" and "files skipped"  statements upped to a higher level than warning

n2y/main.py
sHermanGriffiths: I couldn't find anything that I could make this specific.

xaker00: After some experiments, I figured out that I am wrong. I've seen messages deduplicated when forwarding them to syslog. Apparently syslog will duplicate, but not the logger itself.

Here is my sandbox: https://replit.com/@IlyaSpivakov/SomberAcrobaticDiskdrive

n2y/main.py
xaker00: I'm not sure that I fully understand what you mean. I'm going off of this table

image
https://docs.python.org/3/howto/logging.html#logging-advanced-tutorial

sHermanGriffiths: Because line 113 in main.py (an `info()` log) has to print to stdout, the level of the logger has been lowered to info. Now, since the whole point of making these changes is to make sure logs pertaining to skipping files _don't_ print to stdout, that means they have to be categorized to a logger level lower than info (i.e. debug).

xaker00: Here is what David asked for (issue #13):
> Finally, we should either remove the warning associated with blank files (since this is now expected behavior and there may be many rows with blank slugs) OR we should condense the warnings into a single line of text that says something like "Ignored one or more rows with blank slug column values."

I am not totally sure why a message telling the user to rename a file is a `debug` level message. (I read these particular lines as dealing with duplicates and not blanks)

sHermanGriffiths: While David and I were discussing how to handle duplicate file names, we basically came to the conclusion that we should just skip files that were duplicated in favor of letting the end-users rename them independently instead of using the algorithm in the [dupefileWarning](https://github.com/innolitics/n2y/blob/dupefileWarning/n2y/main.py#:~:text=if%20filename%20in,file_names.append(filename) branch to rename them. Because I figured he would want those warnings added to that condensed final message, not printed immediately, I asked if that were true. He confirmed that my assumption was correct, so I added that functionality here.

xaker00: I'll defer to @johndgiese

sHermanGriffiths: Reason I choose the debug level over the error level was the same reasoning as above.

n2y/main.py
```diff
@@ -61,67 +68,83 @@ def name_column_valid(raw_rows, name_column):
...
+
+                    # sanitize file name just a bit
+                    # maybe use python-slugify in the future?
+                    with open(os.path.join(options.target, f"{filename}.md"), 'w') as f:
+                        f.write('---\n')
+                        f.write(yaml.dump(meta))
+                        f.write('---\n\n')
+                        f.write(markdown)
+                else:
+                    logger.debug('Skipping page "%s" because it is empty.', page_name)
```
johndgiese, line 127: I think it's appropriate that this message is at the "debug" level, since it's usually not interesting and is only useful for debugging a problem.

n2y/main.py
johndgiese: I think this message should be a warning, since we don't typically expect that two pages with the same name will be present. It doesn't feel like it's an error either though, since the software can continue running as expected.

n2y/main.py
```diff
@@ -61,67 +68,83 @@ def name_column_valid(raw_rows, name_column):
...
+    types_skipped = 0
+    prefixes = ("", " & ", ", & ")
+    for key in skips.keys():
+        count = skips[key]
+        if count > 0:
+            msg = msg if key != "duplicate" and types_skipped < 2 else msg.replace(" &", ",")
+            msg += f"{prefixes[types_skipped]}{count} {key}"
+            types_skipped += 1
+
+    msg == "" or logger.error(f"{msg} page(s) skipped")
```
johndgiese: This message should be at the info level, since we will often have many draft articles which will have blank slug column values. Thus, skipping some pages doesn't even warrant an "warning"; everything is working as expected.

n2y/simplify.py
```diff
@@ -45,17 +47,17 @@ def simplify_property(prop):
     elif prop["type"] == "multi_select":
         return [option["name"] for option in prop["multi_select"]]
     elif prop["type"] == "relation":
-        print("WARNING: field type 'relation' is not fully implemented", file=sys.stderr)
+        logger.warning("field type 'relation' is not fully implemented")
```
johndgiese, line 50: Nice! It seems that all of these are correctly set as warnings.

tests/test_converter.py
```diff
@@ -1,6 +1,6 @@
+import pytest
```
johndgiese, line 1: The original position of the `purest` import was correct, since pytest is a third-party module. Here's the PEP8 pattern:

```
stdlib import
stdlib import
stdlib import

third party import
third party import
third party import

internal import
internal import
```

sHermanGriffiths: oh shoot, thought it was stdlib

xaker00 (approved):

Issue #32, opened 2022-06-29
Support Synced Blocks


johndgiese:
See https://developers.notion.com/reference/block#synced-block-blocks

Also see https://github.com/innolitics/n2y/blob/main/n2y/blocks.py#L471-L472

Requirements:

- [ ] The synced block needs to pull in all of the content from the main block
- [ ] Look into what happens if the original block is not shared with the API but the "child" synced block is shared. Add good error messages as appropriate.

Verification Notes:

- [ ] Be sure to update the "All Blocks" end to end test

PR #35, opened 2022-07-06
Implement SyncedBlock

bimbashrestha:
Closes https://github.com/innolitics/n2y/issues/32

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -468,8 +468,9 @@ class LinkPreviewBlock(WarningBlock):
     pass
 
 
-class SyncedBlock(WarningBlock):
-    pass
+class SyncedBlock(Block):
+    def to_pandoc(self):
+        return self.children_to_pandoc()
 
 
 class LinkToPageBlock(WarningBlock):
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -470,6 +470,10 @@ class LinkPreviewBlock(WarningBlock):
 
 class SyncedBlock(Block):
     def to_pandoc(self):
+        synced_block_shared = self.has_children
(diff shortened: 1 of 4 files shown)


tests/test_blocks.py
```diff
@@ -366,3 +366,29 @@ def test_callout():
         '\n'
         'Children\n'
     )
+
+
+def test_synced_block():
```
johndgiese, line 371: I'd split this out into two tests that are more granular, that way the test names (and any failures) can more quickly be honed into.

E.g., `test_synced_block_shared` and `test_scyned_block_unshared`.

n2y/blocks.py
```diff
@@ -468,8 +468,13 @@ class LinkPreviewBlock(WarningBlock):
     pass
 
 
-class SyncedBlock(WarningBlock):
-    pass
+class SyncedBlock(Block):
+    def to_pandoc(self):
+        synced_block_shared = self.has_children
```
johndgiese, line 473: Nice use of an intermediate variable name to explain what's going on.

n2y/blocks.py
```diff
@@ -468,8 +468,13 @@ class LinkPreviewBlock(WarningBlock):
...
+        if not synced_block_shared:
+            logger.warning('Skipping un-shared synced block (%s)', self.notion_url)
```
johndgiese, line 475: 👍 this should definitely be a warning; nice job including the url; it makes it easy to track down and fix this stuff

johndgiese (commented):
@bimbashrestha nice work! I had one small change with the test. Also please add an entry to the change log in the readme; otherwise it looks good! Feel free to merge once both of these are addressed.

PR #39, opened 2022-07-08
Pull in original block contents into synced block

bimbashrestha:
Closes https://github.com/innolitics/n2y/issues/38

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -472,6 +472,8 @@ class SyncedBlock(Block):
     def __init__(self, client, notion_data, page, get_children=True):
         self.original = notion_data[notion_data["type"]]["synced_from"] is None
         super().__init__(client, notion_data, page, get_children=self.original)
+        # Synced blocks will always have children unless not shared
+        # (There will always be at least one UnsupportedBlock child)
         self.shared = self.has_children
         self.children = self._get_synced_block_children()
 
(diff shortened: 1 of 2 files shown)


n2y/blocks.py
```diff
@@ -469,9 +469,21 @@ class LinkPreviewBlock(WarningBlock):
 
 
 class SyncedBlock(Block):
+    def __init__(self, client, notion_data, page, get_children=True):
+        super().__init__(client, notion_data, page, get_children)
```
johndgiese, line 473: Will the `Block.__init__` also call `self.client.get_child_blocks`? If so, since we're over-riding the `self.children` property here, maybe we could pass in `get_children=False` to avoid an unnecessary API call?

n2y/blocks.py
```diff
@@ -469,9 +469,21 @@ class LinkPreviewBlock(WarningBlock):
...
+        self.shared = self.has_children
```
johndgiese, line 474: What happens if you have an empty synced block? Will `has_children` be false in that case? If so, would we be printing a spurious warning in this situation?

johndgiese (commented):
@bimbashrestha I had a few questions. Let me know once you've addressed them all and I can do another review.

n2y/blocks.py
johndgiese: Sweet, thanks for checking! This is good to know. Maybe worth adding a comment about this for our future selves?

n2y/blocks.py
bimbashrestha: Hm, good question.

It's only overriding self.children when `not self.original and self.shared`. So the only time it would be performing an unnecessary API call might be when it's a reference synced block on the same page. I can pass the `self.original` flag into the super constructor and that should remove all redundant API calls

johndgiese (approved):
All of these changes look great! Nice work!

PR #42, opened 2022-07-13
Small refactor

johndgiese:


n2y/config.py
```diff
@@ -0,0 +1,43 @@
+import json
```
bimbashrestha, line 1: Nice! Pulling this out into it's own file makes total sense since it's likely to grow in size.

n2y/utils.py
```diff
@@ -99,5 +99,13 @@ def sanitize_filename(filename):
     return s
 
 
+def id_from_share_link(share_link):
+    hyphens_removed = share_link.replace("-", "")
```
bimbashrestha, line 103: Perhaps we could just call `strip_dashes()` here?

Also it might make sense to rename `strip_dashes()` -> `strip_hyphens()` for consistency.

bimbashrestha (approved):
Nice refactor @johndgiese. Looks good to me! Just pointed out some nit picks.

bimbashrestha (approved):

PR #44, opened 2022-07-27
Set useMaxWidth to False via mermaid config file

bimbashrestha:
This removes white spaces above and below generated diagrams in the final documents.

### DIFF
--- a/n2y/plugins/mermaid.py
+++ b/n2y/plugins/mermaid.py
@@ -2,6 +2,7 @@ import logging
 import os
 import subprocess
 import tempfile
+import json
 
 from pandoc.types import Para, Image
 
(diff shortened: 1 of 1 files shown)


n2y/plugins/mermaid.py
```diff
@@ -32,10 +39,14 @@ def __init__(self, client, page, notion_data, get_children=True):
     def to_pandoc(self):
         temp_fd, temp_filepath = tempfile.mkstemp(suffix=".png")
         os.close(temp_fd)
+        temp_config_fd, temp_config_filepath = tempfile.mkstemp(suffix=".json")
```
johndgiese, line 42: What is the benefit of generating the config file on the fly, vs just including a config file in the python source code?

bimbashrestha: Hm, good question. I think the main benefit would be that it'd be easier to add more config parameters/have dynamic  parameters (derived from cli arguments perhaps) if it's defined as a python dict like this.

johndgiese: That makes sense and seems like a good argument to me!

johndgiese (approved):

Issue #46, opened 2022-08-05
Fix two new bugs


johndgiese:
See https://github.com/innolitics/n2y/commit/789ce5965d0392cb24bf131bc39a55e112d6bf7a

PR #47, opened 2022-08-05
Fix failing end-to-end tests (italicized space bug, quoteblock second paragraph bug)

bimbashrestha:
Closes https://github.com/innolitics/n2y/issues/46

Fixes the two failing asserts in the end-to-end test
1. It looks like Notion encodes spaces in Latin1 (strangely this has only appeared now with italicized words?), so we need to make sure we append Space()s for them like we do for utf-8 spaces. 
2. QuoteBlocks can have children and right now the to_pandoc() method for it doesn't handle that. I made it handle that by extending the ParagraphBlock (because the QuoteBlock is effectively just a ParagraphBlock wrapped in quotes).

### DIFF
--- a/n2y/rich_text.py
+++ b/n2y/rich_text.py
@@ -42,11 +42,13 @@ class RichText:
 
     def plain_text_to_pandoc(self):
         ast = []
-        match = re.findall(r"( +)|(\S+)|(\n+)|(\t+)", self.plain_text)
+        match = re.findall(r"( +)|(\xa0+)|(\S+)|(\n+)|(\t+)", self.plain_text)
 
         for m in match:
-            space, word, newline, tab = m
+            space, latin1_space, word, newline, tab = m
             for _ in range(len(space)):
+                ast.append(Space())
+            for _ in range(len(latin1_space)):
                 ast.append(Space())
             if word:
                 ast.append(Str(word))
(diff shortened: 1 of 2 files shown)


n2y/rich_text.py
```diff
@@ -42,12 +42,14 @@ def to_markdown(self):
 
     def plain_text_to_pandoc(self):
         ast = []
-        match = re.findall(r"( +)|(\S+)|(\n+)|(\t+)", self.plain_text)
+        match = re.findall(r"( +)|(\xa0+)|(\S+)|(\n+)|(\t+)", self.plain_text)
 
         for m in match:
-            space, word, newline, tab = m
+            space, latin1_space, word, newline, tab = m
```
johndgiese, line 48: Are there any other ways to encode a space?

johndgiese (approved):
Nice work! I like how you used the paragraph block

n2y/rich_text.py
bimbashrestha: Good question, I don't think so. I looked through the Latin1 character set (which looks to be what the Notion API supports) and only found two ways to encode a space [1]. There is the default encoding and then this new one (also known as a "non-breaking space"). 

[1]: https://kb.iu.edu/d/aepu

Issue #37, opened 2022-07-08
Add a builtin plugin for supporting footnotes


johndgiese:
We'd like to support Pandoc-style foot notes in our notion articles. E.g., see https://www.notion.so/innolitics/2022-05-29-multiple-function-device-products-policy-and-considerations-32cbd7b904364ca1add119e870148ba7#4fda0a6f04214f19a4f38246ce9bc373

However, if we use n2y to convert this page to markdown the foot notes will be escaped. This is because our current RichText classes and block classes in n2y are not aware of footnotes and as a result, the AST trees that are produced by n2y do note use pandocs footnote nodes. Thus, when Pandoc generates the markdown for the AST, the footnotes are escaped.

To do this property, we'll need to create an n2y plugin that can properly handle foot notes. This will probably involve overriding the pargraph class if it begins with `[NUMBER]: `, e.g., `[4]: This is a foot note definition`. It will also likely involve overriding RichText objects if they make reference to the foot notes. 

See the "Note" data constructor here: https://hackage.haskell.org/package/pandoc-types-1.22.2/docs/Text-Pandoc-Definition.html#t:Inline

Note that in Pandoc's AST, the foot note reference and the content of the footnote are lumped together. Since this data may be separated in the notion pages, we will need to have some data sharing mechanism to collect and then combine the data when the `to_pandoc` calls are run. The proper generic mechanism for this cross-block sharing within plugins isn't fully throught out yet. See https://github.com/innolitics/n2y/blob/main/n2y/notion.py#L87 for a possible starting place.

I think Pandoc's data model won't allow multiple references to the same foot note. I.e., you can't do "Here is a paragraph [^4] asdfasdf [^4]." I _think_ this IS supported in GitHub flavored markdown however. Thus, we should warn users that foot notes are being dropped if this happens. We should also warn users if there is a foot note citation that doesn't have content OR of there is footnote content that isn't referenced.

 Note that one of the end to end tests for the plugins has some footnotes set up in it already.

johndgiese:
This issue is really needed for our website, not for client work, thus do not start on it for now.

PR #48, opened 2022-08-09
Add support for footnotes

bimbashrestha:
Closes https://github.com/innolitics/n2y/issues/37

TODO (will do this once you think this approach looks fine @johndgiese): 
* [x] Periods after the footnote reference are currently gobbled up (easy fix though)
* [x] Add unit tests for footnotes plugin
* [x] Add warning for multiple references to the same footnote
* [x] Add warning if footnote contains no content

### DIFF
--- a/n2y/main.py
+++ b/n2y/main.py
@@ -157,11 +157,23 @@ def main(raw_args, access_token):
     return 0
 
 
+def get_database_pages_and_prefetch_blocks(database):
+    pages = []
+    for page in database.children:
+        # We're doing this because plugins may need to do some preprocessing
+        # on blocks before they're rendred by to_pandoc(). Because page.block
+        # is a lazy-loaded property, we need to pre-fetch it to make sure all
+        # block contructors are called before all block to_pandoc() calls.
+        _ = page.block
+        pages.append(page)
+    return pages
+
+
 def export_database_as_markdown_files(database, options):
     os.makedirs(options.output, exist_ok=True)
     seen_file_names = set()
     counts = {'unnamed': 0, 'duplicate': 0}
-    for page in database.children:
+    for page in get_database_pages_and_prefetch_blocks(database):
         if page.filename:
             if page.filename not in seen_file_names:
                 seen_file_names.add(page.filename)
--- a/n2y/main.py
+++ b/n2y/main.py
@@ -157,23 +157,11 @@ def main(raw_args, access_token):
     return 0
 
 
-def get_database_pages_and_prefetch_blocks(database):
-    pages = []
-    for page in database.children:
-        # We're doing this because plugins may need to do some preprocessing
-        # on blocks before they're rendred by to_pandoc(). Because page.block
-        # is a lazy-loaded property, we need to pre-fetch it to make sure all
-        # block contructors are called before all block to_pandoc() calls.
-        _ = page.block
-        pages.append(page)
-    return pages
-
-
 def export_database_as_markdown_files(database, options):
     os.makedirs(options.output, exist_ok=True)
     seen_file_names = set()
     counts = {'unnamed': 0, 'duplicate': 0}
-    for page in get_database_pages_and_prefetch_blocks(database):
+    for page in database.children:
         if page.filename:
             if page.filename not in seen_file_names:
                 seen_file_names.add(page.filename)
--- a/n2y/plugins/footnotes.py
+++ b/n2y/plugins/footnotes.py
@@ -4,6 +4,7 @@ from pandoc.types import Note, Str, Para
 
 from n2y.rich_text import TextRichText
 from n2y.blocks import ParagraphBlock
+from n2y.errors import UseNextClass
 
 
 plugin_data_key = "n2y.plugins.footnotes"
@@ -12,16 +13,18 @@ plugin_data_key = "n2y.plugins.footnotes"
 class ParagraphWithFootnoteBlock(ParagraphBlock):
     def __init__(self, client, notion_data, page, get_children=True):
         super().__init__(client, notion_data, page, get_children)
-        self._attach_footnote_data_if_exists()
+        if self._is_footnote():
+            self._attach_footnote_data()
+        else:
+            raise UseNextClass()
 
     def to_pandoc(self):
         return None if self._is_footnote() else super().to_pandoc()
 
-    def _attach_footnote_data_if_exists(self):
-        if self._is_footnote():
-            if plugin_data_key not in self.client.plugin_data:
-                self.client.plugin_data[plugin_data_key] = {}
-            self.client.plugin_data[plugin_data_key][self._footnote()] = self._footnote_ast()
+    def _attach_footnote_data(self):
+        if plugin_data_key not in self.client.plugin_data:
+            self.client.plugin_data[plugin_data_key] = {}
+        self.client.plugin_data[plugin_data_key][self._footnote()] = self._footnote_ast()
 
     def _is_footnote(self):
         return self._footnote() is not None
--- a/n2y/plugins/footnotes.py
+++ b/n2y/plugins/footnotes.py
@@ -45,7 +45,13 @@ class ParagraphWithFootnoteBlock(ParagraphBlock):
 
     def _footnote_ast(self):
         ast = super().to_pandoc()
-        return Para(ast[0][2:]) if not isinstance(ast, list) else [Para(ast[0][0][2:])] + ast[1:]
+        if isinstance(ast, list):
+            first_paragraph_footnote_stripped = Para(ast[0][0][2:])
+            remaining_paragraphs = ast[1:]
+            return [first_paragraph_footnote_stripped] + remaining_paragraphs
+        else:
+            paragraph_footnote_stripped = Para(ast[0][2:])
+            return paragraph_footnote_stripped
 
 
 class TextRichTextWithFootnoteRef(TextRichText):
--- a/n2y/plugins/footnotes.py
+++ b/n2y/plugins/footnotes.py
@@ -1,0 +1,58 @@
+import re
+
+from pandoc.types import Note, Str, Para
+
+from n2y.rich_text import TextRichText
+from n2y.blocks import ParagraphBlock
+
+
+class ParagraphWithFootnoteBlock(ParagraphBlock):
+    def __init__(self, client, notion_data, page, get_children=True):
+        super().__init__(client, notion_data, page, get_children)
+        self._attach_footnote_data_if_exists()
+
+    def to_pandoc(self):
+        return None if self._is_footnote() else super().to_pandoc()
+
+    def _attach_footnote_data_if_exists(self):
+        if self._is_footnote():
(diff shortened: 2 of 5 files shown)


johndgiese (commented):
@bimbashrestha why is it necessary to take second pass through the rendering here? It seems like we already kind of have two passes. In the first pass, we're constructing all of the notion objects (e.g., the subclasses of Block, RichTextArray, Mention, etc.). In the second pass we run `to_pandoc` on everything.

Couldn't we collect all of the footnotes in the constructor phase, say by attaching data to the plugin_data dict on the client (see https://github.com/innolitics/n2y/blob/main/n2y/notion.py#L94) and then using this to look up the foot notes after? Alternatively, it could make name spacing easier of the `plugin_data` were on the associated Page object.

Also, if we did move forward with a two-pass approach, I think we would want to only do it if there were particular block instances that needed it instead of only relying on block classes that need it. This is because even if you have a plugin with a block that sometimes requires a second pass, I bet most of the time you wouldn't need it. I realize this would be more complicated to do, but again, maybe we can use the constructor phase as the first pass and the to_pandoc as the second pass, and thus avoid this complexity.

I very well may be missing some reason as to why the two pass system you added is necessary though, in which case we should definitely discuss.

bimbashrestha:
Hey @johndgiese good question. I think I might have misunderstood the data flow. Let me take a second stab at this today. 

I suspect you're right and it's possible to do without the second pass logic I have now.

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,58 @@
...
+        pandoc_ast = []
+        for token in super().to_pandoc():
+            if not isinstance(token, Str):
+                pandoc_ast.append(token)
+                continue
+            refs = re.findall(r"\[\^(\d+)\]", token[0])
+            if len(refs) != 1:
+                pandoc_ast.append(token)
+                continue
+            block = self.client.plugin_data["footnotes"][refs[0]]
```
johndgiese, line 49: What would happen if there are multiple pages with foot notes?

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,58 @@
...
+        super().__init__(client, notion_data, page, get_children)
+        self._attach_footnote_data_if_exists()
+
+    def to_pandoc(self):
+        return None if self._is_footnote() else super().to_pandoc()
+
+    def _attach_footnote_data_if_exists(self):
+        if self._is_footnote():
+            if "footnotes" not in self.client.plugin_data:
+                self.client.plugin_data["footnotes"] = {}
```
johndgiese, line 20: You may want to use `n2y.plugins.footnotes` instead of `footnotes` here, to reduce the chances of name clashes.

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,58 @@
...
+            self.client.plugin_data["footnotes"][self._footnote()] = self._footnote_ast()
```
johndgiese, line 21: I'd show a warning if the same footnote shows up twice in the same file (and be sure to include a link to the second footnote block)

n2y/plugins/footnotes.py
johndgiese, line 49: I'd also handle the case when a foot note is missing and show good warnings as appropriate (but still do your best to render something reasonable in the file).

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,58 @@
...
+                continue
+            refs = re.findall(r"\[\^(\d+)\]", token[0])
+            if len(refs) != 1:
+                pandoc_ast.append(token)
+                continue
+            block = self.client.plugin_data["footnotes"][refs[0]]
+            footnote = Note(block) if isinstance(block, list) else Note([block])
+            pandoc_ast.append(footnote)
+        return pandoc_ast
+
```
johndgiese, line 53: I don't know how exactly you'd do this, but it may be good to add a warning for foot note content that isn't used.

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,58 @@
...
+from pandoc.types import Note, Str, Para
+
+from n2y.rich_text import TextRichText
+from n2y.blocks import ParagraphBlock
+
+
+class ParagraphWithFootnoteBlock(ParagraphBlock):
+    def __init__(self, client, notion_data, page, get_children=True):
+        super().__init__(client, notion_data, page, get_children)
+        self._attach_footnote_data_if_exists()
```
johndgiese, line 12: If the foot note data doesn't exist, I'd raise a `UseNextBlock` exception so that other plugins can run if appropriate.

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,58 @@
...
+            return None
+        return footnotes[0]
+
+    def _footnote_ast(self):
+        ast = super().to_pandoc()
+        return Para(ast[0][2:]) if not isinstance(ast, list) else [Para(ast[0][0][2:])] + ast[1:]
+
+
+class TextRichTextWithFootnoteRef(TextRichText):
+    def to_pandoc(self):
```
johndgiese, line 39: I'd throw a UseNextBlock here as well if we don't actually have a foot note

johndgiese: Actually, to do this you'd need to check if rich text is a foot note or not in the constructor instead of here in `to_pandoc`. If that's easy to do, I'd do it.

n2y/plugins/footnotes.py
bimbashrestha: Good question. I don't think what I have right now will work for multiple pages. Specifically, I don't think it will work when we're trying to export a database of pages with footnotes because blocks inside pages (from what i can tell) are lazy-instansiated. 

I'll address this

n2y/main.py
```diff
@@ -157,11 +157,23 @@ def main(raw_args, access_token):
     return 0
 
 
+def get_database_pages_and_prefetch_blocks(database):
```
bimbashrestha, line 160: Let me know what you think about this @johndgiese. I could just load the block in the Page constructor but I figured there must be a reason why we're lazy-loading it (hence the pre-fetching).

n2y/plugins/footnotes.py
bimbashrestha: Hmm, I suppose we could add a flag per footnote in the plugin_data dictionary and then check to see if they've all been set at the end. Do you think this is worth doing? I definitely see the value a warning would provide but unused footnotes are also relatively harmless?

n2y/plugins/footnotes.py
bimbashrestha: BTW, I addressed this by pre-fetching the blocks inside pages. See my comment below. I'd be curious what you think.

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ParagraphWithFootnoteBlock(ParagraphBlock):
+    def __init__(self, client, notion_data, page, get_children=True):
+        super().__init__(client, notion_data, page, get_children)
+        if self._is_footnote():
+            self._attach_footnote_data()
+        else:
+            raise UseNextClass()
+
+    def to_pandoc(self):
```
johndgiese, line 24: I think we can just use the default `to_pandoc` now that the constructor will raise a `UseNextClass` exception.

johndgiese: Ahh, actually, we will just always return None here

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,88 @@
...
+    def _footnote(self):
+        first_str = self.rich_text.to_plain_text().split(" ")[0]
+        footnotes = re.findall(r"\[(\d+)\]:", first_str)
+        if len(footnotes) != 1:
+            return None
+        return footnotes[0]
+
+    def _footnote_ast(self):
+        ast = super().to_pandoc()
+        return Para(ast[0][2:]) if not isinstance(ast, list) else [Para(ast[0][0][2:])] + ast[1:]
```
johndgiese, line 48: This line is kind of hard to follow. Maybe you could split it out a bit and/or add a comment?

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,88 @@
...
+            raise UseNextClass()
+
+    def to_pandoc(self):
+        return None if self._is_footnote() else super().to_pandoc()
+
+    def _attach_footnote_data(self):
+        if plugin_data_key not in self.client.plugin_data:
+            self.client.plugin_data[plugin_data_key] = {}
+        if self._footnote() not in self.client.plugin_data[plugin_data_key]:
+            self.client.plugin_data[plugin_data_key][self._footnote()] = self._footnote_ast()
```
johndgiese, line 31: Is the result of `self._footnote()` just the number of the foot note? If so, it seems like you won't be able to analyze multple pages that have footnotes since they'll both re-use the same numbers.

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,106 @@
...
+    def to_pandoc(self):
+        pandoc_ast = []
+        for token in super().to_pandoc():
+            ref = self._footnote_from_token(token)
+            if ref is None:
+                pandoc_ast.append(token)
+                continue
+            if ref not in self.block.page.plugin_data[plugin_data_key]:
+                pandoc_ast.append(token)
+                logger.warning('Missing footnote "[%s]". Rendering as plain text', ref)
```
johndgiese, line 78: I'd add a notion url to the block that this occurs in, so someone can easily locate where the issue occurred

n2y/plugins/footnotes.py
johndgiese: I agree, this doesn't seem worth doing, especially since we'd have to make a custom page object for just this I think ....

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,106 @@
...
+    def to_pandoc(self):
+        return None
+
+    def _attach_footnote_data(self):
+        if plugin_data_key not in self.page.plugin_data:
+            self.page.plugin_data[plugin_data_key] = {}
+        if self._footnote() not in self.page.plugin_data[plugin_data_key]:
+            self.page.plugin_data[plugin_data_key][self._footnote()] = self._footnote_ast()
+            if self._footnote_ast_empty():
+                logger.warning('Empty footnote "[%s]"', self._footnote())
```
johndgiese, line 33: Ditto about the notion url

n2y/plugins/footnotes.py
```diff
@@ -0,0 +1,106 @@
...
+        else:
+            msg = 'Multiple footnotes for "[%s]". Skipping latest (%s)'
```
johndgiese, line 35: I think we haven't been using sentences for these (although I may be wrong)

```suggestion
            msg = 'Multiple footnotes for "[%s]", skipping latest (%s)'
```

johndgiese (approved):
Nice work! Please address all the comments and also be sure you can create markdown files for our website for the guidance docs

PR #49, opened 2022-08-10
Add linked block headers plugin

johndgiese:
Closes https://github.com/innolitics/n2y/issues/50

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,10 @@ This plugin assumes that the `mmdc` mermaid commandline tool is available, and w
 
 If there are errors with the mermaid syntax, it is treated as a normal codeblock and the warning is logged.
 
+### Linked Header Blocks
+
+Replace headers with links back to the originating notion block.
+
 ## Architecture
 
 N2y's architecture is divided into four main steps:
@@ -173,6 +177,7 @@ Here are some features we're planning to add in the future:
 
 - Add support for dumping the notion urls using `--url-property`.
 - Add support for all types of rollups (including arrays of other values)
+- Add plugin for creating headers that link back to the notion blocks
 
 ### v0.4.2
 
--- a/n2y/plugins/mermaid.py
+++ b/n2y/plugins/mermaid.py
@@ -33,7 +33,8 @@ class MermaidFencedCodeBlock(FencedCodeBlock):
         temp_fd, temp_filepath = tempfile.mkstemp(suffix=".png")
         os.close(temp_fd)
         try:
-            diagram_as_bytes = self.rich_text.to_plain_text().encode()
+            diagram_as_text = self.rich_text.to_plain_text()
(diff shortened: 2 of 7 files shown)


tests/test_blocks.py
```diff
@@ -21,16 +21,16 @@
 )
 
 
-def generate_block(notion_block):
+def generate_block(notion_block, plugins=None):
```
johndgiese, line 24: 👍 nice

n2y/plugins/mermaid.py
```diff
@@ -43,7 +43,8 @@ def to_pandoc(self):
         os.write(temp_config_fd, json.dumps(mermaid_config).encode("utf-8"))
         os.close(temp_config_fd)
         try:
-            diagram_as_bytes = self.rich_text.to_plain_text().encode()
```
johndgiese, line 46: Nice cleanup!

README.md
```diff
@@ -137,6 +137,10 @@ This plugin assumes that the `mmdc` mermaid commandline tool is available, and w
 
 If there are errors with the mermaid syntax, it is treated as a normal codeblock and the warning is logged.
 
+### Linked Header Blocks
+
+Replace headers with links back to the originating notion block.
```
johndgiese, line 142: This makes it sound like the headers are no longer headers as they're replaced by links. Perhaps say "Make headers have links back to the originating notion block."

tests/test_linkedheaders.py
```diff
@@ -0,0 +1,53 @@
...
+            )
+        ],
+    )
+
+
+def test_heading_1():
+    notion_block = mock_block("heading_1", {"rich_text": [mock_rich_text("Heading One")]})
+    pandoc_ast, markdown = process_block(notion_block, ["n2y.plugins.linkedheaders"])
+    assert pandoc_ast == mock_header_ast(1, "One", notion_block)
+    assert markdown == f'# [Heading One](#{strip_hyphens(notion_block["id"])})\n'
```
johndgiese, line 32: What's the benefit of using what you've got over this:

```
f'# [Heading One]({notion_block["notion_url"]})\n'
```

(or equivalent)?

johndgiese (commented):
Hey @bimbashrestha I can't "approve" since I created the PR, but this looks good! I had a few small suggestions, but overall seems solid!

tests/test_linkedheaders.py
bimbashrestha: Ah, I like that. I didn't see "url" in the mocked notion block dict but perhaps we should add it in?

johndgiese: Ahh, I thought the url was included in the data from Notion. If it is and it's missing, I think we should add it. If it's not included in the payload from Notion, then we shouldn't and what you have now is great. I definitely want the mocks to match the notion api as best as possible.

PR #55, opened 2022-08-11
Add block references

johndgiese:
This should be self explanatory; I needed this for a plugin I am writing

### DIFF
--- a/n2y/rich_text.py
+++ b/n2y/rich_text.py
@@ -17,10 +17,15 @@ class RichText:
     """
     A sequence of text that contains the same styling throughout. It also may
     contain an @-mention or an equation (both of which may be styled as well).
+
+    Contains a reference to the block that the rich text is contained in, if
+    any. If the block reference is None, then the rich text is perhaps used in
+    a property_value or somewhere else.
     """
 
-    def __init__(self, client, notion_data):
+    def __init__(self, client, notion_data, block=None):
         self.client = client
+        self.block = block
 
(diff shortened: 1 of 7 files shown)


n2y/rich_text.py
```diff
@@ -152,31 +158,6 @@ def to_pandoc(self):
             return annotated_ast
 
 
-def mock_notion_annotations(annotations=None):
```
johndgiese, line 155: I had actually pulled these two mocks out since they're used in the core n2y code; now that you moved the other ones over, I moved it into the mocks file.

bimbashrestha:
This looks good to me @johndgiese as soon as CI is passing (looks like it's just flake8 to me).

n2y/rich_text.py
```diff
@@ -134,7 +135,7 @@ def to_pandoc(self):
 class TextRichText(RichText):
     @classmethod
     def from_plain_text(klass, client, string, block=None):
-        notion_data = mock_notion_rich_text(string)
+        notion_data = mock_rich_text(string)
```
bimbashrestha, line 138: Surprised our tests didn't catch this. Guess flake8 did so all is well. Nice!

bimbashrestha (approved):
Looks good to me!

n2y/rich_text.py
johndgiese: I'm not sure where `from_plain_text` is used. I think it may only be in other plugins.

PR #58, opened 2022-08-13
Add support for generating HTML pages

johndgiese:


johndgiese:
FYI @bimbashrestha I needed to merge this in, but please review anyway and we can address changes after the fact

Issue #25, opened 2022-05-09
Implement support for remaining block types


johndgiese:
- [x] Review the blocks code here https://github.com/innolitics/n2y/blob/main/n2y/blocks.py#L457-L491
- [x] Update the README table to add any relevant notes for the various block types
- [x] Every block type should be in the table
- [x] Only add information that is not obvious; e.g., if an paragraph block just makes a paragraph in markdown, we don't need to add any comment about it.
- [x] For blocks that haven't yet been implemented, write out how you think they _should_ work in the readme table; e.g., the Pdf block can download the PDF to the "media root" and add a link in the GitHub to the downloaded PDF file.
- [x] Assign a PR to @bimbashrestha for him to review. The PR should ONLY contain changes to the README at this point
- [x] Once Bimba approves the direction, implement the proposed changes.

sHermanGriffiths:
Addressed with PR #65

PR #63, opened 2022-08-24
Add all blocks to the readme table

sHermanGriffiths:


README.md
```diff
@@ -64,11 +64,47 @@ At the core of n2y are a set of python classes that represent the various parts
 
 | Notion Object Type | Description |
 | --- | --- |
+| Block | A bit of content within a Page |
```
johndgiese, line 67: @sHermanGriffiths this table, which includes the Notion Object Types, should _not_ include all of the individual block types. There is a table further down in the readme for the block types that contains some of the block types already.

Also, the descriptions should be empty unless there is something non-obvious about the behavior. Thus, comments like "A notion block of an image" should be omitted. This was the intent of the GitHub issue that stated "Only add information that is not obvious; e.g., if an paragraph block just makes a paragraph in markdown, we don't need to add any comment about it.". I think this will make the table more useful if it only contains the non-obvious stuff.

README.md
```diff
@@ -64,11 +64,47 @@ At the core of n2y are a set of python classes that represent the various parts
...
+| BreadcrumbBlock | A Notion block which denotes the current page location |
+| UnsupportedBlock | A Notion block that isn't supported by the program yet |
+| TemplateBlock | A template for a new page |
+| WarningBlock | A block that logs a warning for unsupported blocks |
+| ChildDatabaseBlock | A block that represents a database that is the child of the current page |
+| EmbedBlock | A block which embeds one of a selection of features into the page |
+| VideoBlock | A block that houses a video |
+| PdfBlock | A block that houses a PDF |
+| ChildrenPassThroughBlock | A block that passes along the children of it's parent |
+| ColumnBlock | A block that apears on the page as a column |
```
johndgiese, line 99: The description of the ColumnBlock should explain how we convert the column blocks into markdown. In particular, we ignore the columns and just display the content in each column one after the other in the markdown, since there is no way to support columns in markdown. This is a good example of the non-obvious sort of content that should go into this table.

johndgiese (changes requested):
@sHermanGriffiths I think there was some confusion regarding the tasks in https://github.com/innolitics/n2y/issues/25

Please read the issue more carefully next time, as I believe some of the instructions you didn't follow seem pretty clear.

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
 
 | Class Name | Noteworthy Behavior |
 | --- | --- |
-| BookmarkBlock | |
+| BookmarkBlock | Converts visual bookmark into plain text link in markdown, using the caption as the link text. |
+| BreadcrumbBlock | Ignore |
```
johndgiese, line 103:
```suggestion
| BreadcrumbBlock | These blocks are ignored |
```

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
+| SyncedBlock | Transcribe the contents of the synced block at the time it was constructed |
+| TableBlock | |
+| TableOfContentsBlock | transcribes the contenets of the Table of contents block into page links |
+| TemplateBlock | Ignore |
 | ToDoItemBlock | |
-| ToggleBlock | Convert the toggles into a bulleted list. |
-| ColumnBlock | |
-| ColumnListBlock | Columns are flattened and displayed sequentially. |
+| ToggleBlock | Convert toggles into a details and summary tag. |
+| VideoBlock | Acts the same way ass the Image block |
```
johndgiese, line 131:
```suggestion
| VideoBlock | Acts the same way as the Image block |
```

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
 | RowBlock | |
+| SyncedBlock | Transcribe the contents of the synced block at the time it was constructed |
+| TableBlock | |
+| TableOfContentsBlock | transcribes the contenets of the Table of contents block into page links |
+| TemplateBlock | Ignore |
 | ToDoItemBlock | |
-| ToggleBlock | Convert the toggles into a bulleted list. |
-| ColumnBlock | |
-| ColumnListBlock | Columns are flattened and displayed sequentially. |
+| ToggleBlock | Convert toggles into a details and summary tag. |
```
johndgiese, line 130: I think the current default functionality for the ToggleBlock is appropriate. Users can use a plugin if they want to produced details and summary tags. Keep in mind we are often exporting to non-HTML formats, like PDF or Word documents.

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
 | ParagraphBlock | |
+| PdfBlock | Acts the same way ass the Image block |
 | QuoteBlock | |
-| EquationBlock | Converted to "display math" using LaTeX; see the [pandoc](https://pandoc.org/MANUAL.html#math) documentation. |
-| TableBlock | |
 | RowBlock | |
+| SyncedBlock | Transcribe the contents of the synced block at the time it was constructed |
+| TableBlock | |
+| TableOfContentsBlock | transcribes the contenets of the Table of contents block into page links |
+| TemplateBlock | Ignore |
```
johndgiese, line 128: Ditto

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
 | NumberedListItemBlock | |
 | ParagraphBlock | |
+| PdfBlock | Acts the same way ass the Image block |
 | QuoteBlock | |
-| EquationBlock | Converted to "display math" using LaTeX; see the [pandoc](https://pandoc.org/MANUAL.html#math) documentation. |
-| TableBlock | |
 | RowBlock | |
+| SyncedBlock | Transcribe the contents of the synced block at the time it was constructed |
+| TableBlock | |
+| TableOfContentsBlock | transcribes the contenets of the Table of contents block into page links |
```
johndgiese, line 127: I think we can ignore the table of contents blocks for now.

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
+| FileBlock | Acts the same way as the ImageBlock, except that in the documents it only ever shows the URL. |
 | HeadingOneBlock | |
 | HeadingTwoBlock | |
 | HeadingThreeBlock | |
 | ImageBlock | It uses the URL for external images, but downloads uploaded images to the `MEDIA_ROOT` and replaces the path with a relative url based off of `MEDIA_URL`. The "caption" is used for the alt text. |
-| FileBlock | Acts the same way as the ImageBlock, except that in the documents it only ever shows the URL. |
+| LinkToPageBlock | Transcribes the block into a plain text link |
 | NumberedListItemBlock | |
 | ParagraphBlock | |
+| PdfBlock | Acts the same way ass the Image block |
```
johndgiese, line 122:
```suggestion
| PdfBlock | Acts the same way as the Image block |
```

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
+| EmbedBlock | Ignore |
+| EquationBlock | Converted to "display math" using LaTeX; see the [pandoc](https://pandoc.org/MANUAL.html#math) documentation. |
+| FencedCodeBlock | |
+| FileBlock | Acts the same way as the ImageBlock, except that in the documents it only ever shows the URL. |
 | HeadingOneBlock | |
 | HeadingTwoBlock | |
 | HeadingThreeBlock | |
 | ImageBlock | It uses the URL for external images, but downloads uploaded images to the `MEDIA_ROOT` and replaces the path with a relative url based off of `MEDIA_URL`. The "caption" is used for the alt text. |
-| FileBlock | Acts the same way as the ImageBlock, except that in the documents it only ever shows the URL. |
+| LinkToPageBlock | Transcribes the block into a plain text link |
```
johndgiese, line 119: 👍

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
 | Class Name | Noteworthy Behavior |
 | --- | --- |
-| BookmarkBlock | |
+| BookmarkBlock | Converts visual bookmark into plain text link in markdown, using the caption as the link text. |
+| BreadcrumbBlock | Ignore |
 | BulletedListItemBlock | |
 | CalloutBlock | The content of the callout block is extracted, but the emoji and background color are ignored. |
-| ChildPageBlock | |
-| FencedCodeBlock | |
+| ChildDatabaseBlock | Converts database to markdown in different file in directory and writes a relative link to the markdown|
```
johndgiese, line 106: The difficulty with exporting the child database to markdown is that there's no way to represent a database as a single markdown file. Depending on the selected output format, a database is either a single YAML file or many markdown files.

For now, I propose we just ignore child databases.

README.md
```diff
@@ -99,27 +99,36 @@ Here are the default block classes that can be extended:
...
+| ChildPageBlock | Converts Child Page to markdown. |
```
johndgiese, line 107: Depending on the selected output format Command-line argument, we may or many not be exporting the child pages into markdown. In fact, currently, we don't ever export child page objects. In the future, we will have a recursive export format option that will recursively go through a page and export everything.

johndgiese: For now, I propose we just ignore child pages. In the future, I think the solution you have outlined is a good one.

johndgiese (approved):

PR #65, opened 2022-08-29
Issue25

sHermanGriffiths:
Hey, I added a new way of styling the captions. It occurred to me that using the image caption as it's alt isn't the functionality that I'd expect as a user. Same with using the caption as link text. What if the caption goes straight into describing a video? That would look kind of awkward on the page. This new method would use the file url as the link text and img alt by default and render the caption underneath that. Right now it's rendering as a grid in order to distinguish it from normal text but it can also just be rendered as normal text (the code for that is just commented out right now).


n2y/blocks.py
```diff
@@ -2,11 +2,11 @@
 import logging
 from urllib.parse import urljoin
 
-from pandoc.types import (
+from pandoc.types import (  # noqa: F401
```
johndgiese, line 5: Why is F401 necessary here?

n2y/blocks.py
```diff
@@ -311,11 +311,12 @@ def to_pandoc(self):
             url = self.file.url
         elif self.file.type == "file":
             url = self.client.download_file(self.file.url, self.page)
+        content_ast = [Link(('', [], []), [Str(url)], (url, ''))]
         if self.caption:
             caption_ast = self.caption.to_pandoc()
-        else:
-            caption_ast = [Str(url)]
-        return Para([Link(('', [], []), caption_ast, (url, ''))])
+            # content.extend([LineBreak(), *caption_ast])
```
johndgiese, line 317: leftover comment?

n2y/blocks.py
```diff
@@ -388,6 +392,44 @@ def to_pandoc(self):
...
+                ('', [], []),
+                RowHeadColumns(0), [],
+                [Row(('', [], []), cells)])],
+            TableFoot(('', [], []), [])
+        )
+        return table
+
+
+class ColumnBlock(Block):
+    def __init__(self, client, notion_data, page, get_children=True):
```
johndgiese, line 417: Why do you override the `__init__` here, since it doesn't seem to do anything different than `Block.__init__`.?

tests/test_blocks.py
```diff
@@ -211,8 +211,7 @@ def test_image_internal_with_caption(mock_download):
     mock_download.return_value = 'image.png'
     pandoc_ast, markdown = process_block(notion_block)
     assert pandoc_ast == Para([
-        Image(('', [], []), [Str('test'), Space(), Str('image')], ('image.png', ''))
-    ])
+        Image(('', [], []), [Str('test'), Space(), Str('image')], ('image.png', ''))])
```
johndgiese, line 214: In this repo, the convention is to keep the closing brackets on its own line as it was prior to this change.

tests/test_blocks.py
```diff
@@ -411,12 +412,12 @@ def test_scyned_block_unshared():
     assert unshared_reference_markdown == ""
 
 
-def test_column_block():
-    column_block = mock_block("column", {}, has_children=True)
-    children = [mock_paragraph_block([("child", [])])]
-    pandoc_ast, markdown = process_parent_block(column_block, children)
-    assert pandoc_ast == [Para([Str('child')])]
-    assert markdown == "child\n"
+# def test_column_block():
```
johndgiese, line 415: Why is this commented out?

tests/test_end_to_end.py
```diff
@@ -181,6 +181,16 @@ def test_all_blocks_page_to_markdown(tmp_path):
     lines = document_as_markdown.split('\n')
     metadata = parse_yaml_front_matter(document_as_markdown)
     assert metadata['title'] == 'All Blocks Test Page'
+    column_string = (
+        '<table><tbody><tr class="odd"><td>Column 1<table>'
+        '<tbody><tr class="odd"><td>Column 1.1</td><td>Column 1.2</td></tr>'
+        '</tbody></table></td><td>Column 2</td></tr></tbody></table>')
```
johndgiese, line 187: Ditto about the trailing close parenthesis

tests/test_end_to_end.py
```diff
@@ -201,14 +211,14 @@ def test_all_blocks_page_to_markdown(tmp_path):
...
-    assert "Column 1" in lines and "Column 2" in lines
-    assert "Column 1.1" in lines and "Column 1.2" in lines
+    assert all(column_strings_in_lines) or (column_string in lines)
 
     # a bookmark with a caption and without
     assert "<https://innolitics.com>" in lines
     assert "[Bookmark caption](https://innolitics.com)" in lines
 
     # the word "caption" is bolded
+    print(lines)
```
johndgiese, line 221: Leftover print statement?

johndgiese (changes requested):
@sHermanGriffiths I did a quick review of some obvious items. I think a few of these should have been caught prior to requesting the review.

n2y/blocks.py
sHermanGriffiths: the `Space` ast class is unused but the test suite throws this error if it's not imported:
```
n2y/blocks.py:423: in to_pandoc
    plain_children = eval(plain_children_string)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   NameError: name 'Space' is not defined

<string>:1: NameError
```

n2y/blocks.py
sHermanGriffiths: yes

n2y/blocks.py
```diff
@@ -388,6 +392,44 @@ def to_pandoc(self):
...
+        super().__init__(client, notion_data, page, get_children)
+
+    def to_pandoc(self):
+        string_of_children = repr(self.children_to_pandoc())
```
johndgiese, line 421: Are these lines trying to replace each `Para` call with a `Plain`?

If so, why are we trying to do that? Is it because we can't support paragraphs within a table in pandoc's data model?

If we can't support paragraphs in a table, then maybe we should reconsider upgrading the column block like this, since I think Paragraph blocks in a ColumnBlock is going to be pretty common.

I am surprised that we would need to use an `eval` call here? (In general, `eval` is very dangerous and should be avoided.) I bet there would be a way to swap the Para with Plain without using eval.

n2y/blocks.py
sHermanGriffiths: left over from print statements

n2y/blocks.py
sHermanGriffiths: I ran into an error earlier but I can't replicate it now so the `Para` wasn't the issue. I had no clue about `eval`, I'll look into that.

tests/test_blocks.py
sHermanGriffiths: the `ColumnBlock` now returns a `Cell` instead of a `Para` so it can't be fed into `pandoc.write` like it is in the `pandoc_write_or_log_errors` function (via the the `process_parent_block` function)

sHermanGriffiths: If you want still want a test for the column block, I can pare it down to only testing that it returns the expected pandoc...

n2y/blocks.py
sHermanGriffiths: ast.literal_eval is safer, right?

n2y/blocks.py
sHermanGriffiths: Got rid of the eval statement so it's now unnecessary

johndgiese: @sHermanGriffiths I'm not sure if you've pushed up your most recent code. I still see the "nova: F401" in the diff.

n2y/blocks.py
```diff
@@ -519,6 +572,19 @@ class LinkToPageBlock(WarningBlock):
...
+    header_cell_args = [('', [], []), AlignDefault(), RowSpan(1), ColSpan(1), [Plain(content_ast)]]
+    body_cell_args = [('', [], []), AlignDefault(), RowSpan(1), ColSpan(1), [Plain(caption_ast)]]
+    body_row = Row(('', [], []), [Cell(*body_cell_args)])
+    return Table(
+        ('', [], []),
+        Caption(None, []),
+        [(AlignDefault(), ColWidthDefault())],
+        TableHead(('', [], []), [Row(('', [], []), [Cell(*header_cell_args)])]),
+        [TableBody(('', [], []), RowHeadColumns(0), [], [body_row])],
+        TableFoot(('', [], []), []))
```
johndgiese, line 585: Closing parens. We will update flake8 to call out these sometime soon.

tests/test_blocks.py
```diff
@@ -429,5 +439,13 @@ def test_column_list_block(mock_get_child_notion_blocks):
...
-    assert markdown == "child1\n\nchild2\n"
+    cell1 = Cell(('', [], []), AlignDefault(), RowSpan(1), ColSpan(1), [Para([Str('child1')])])
+    cell2 = Cell(('', [], []), AlignDefault(), RowSpan(1), ColSpan(1), [Para([Str('child2')])])
+    assert pandoc_ast == Table(
+        ('', [], []),
+        Caption(None, []),
+        [(AlignDefault(), ColWidthDefault()), (AlignDefault(), ColWidthDefault())],
+        TableHead(('', [], []), []),
+        [TableBody(('', [], []), RowHeadColumns(0), [], [Row(('', [], []), [cell1, cell2])])],
+        TableFoot(('', [], []), []))
```
johndgiese, line 450:
```suggestion
        TableFoot(('', [], []), [])
    )
```

johndgiese (approved):
Nice work @sHermanGriffiths !

PR #68, opened 2022-08-31
Add support for mentioning falsey user name

sHermanGriffiths:


johndgiese (approved):
Nice work tracking this down! I'd add "Fixed bug with user mentions for users that have no name (e.g., integration users)." to the readme.

PR #69, opened 2022-09-09
v.1.0 ready for your review

idolcemia:
Everything is written, but I'm not sure it's working. 

Actually, I am pretty sure....sure there'll be many changes. 

I wrote a test suite (such as it is) which I included it in this commit, as an FYI.  

I wrote the test suite because I was unable to test the plugin as an actual plugin.  I spent some time trying to solve what I expected to be a simple problem, but gave up to focus on the core of the assignment.  

I discovered an upside to not being able to test the plugin in situ.  The process of creating the test suite gave me a much deeper understanding of n2y than I would have otherwise had. 

Thanks you for the opportunity to work on this project.   It sent me in a lot of directions, or so it felt.  And I enjoyed getting into the code base very much.  I was mightily impressed by the level of work.


n2y/notion_mocks.py
```diff
@@ -110,6 +110,15 @@ def mock_file(url):
     }
 
 
+def mock_link_to_page_block(linked_page_id):
```
johndgiese, line 113: A `mock_link_to_page_block` function is probably not necessary, since you can just use `mock_block` with the appropriate values for the `content` argument. That being said, if you find yourself using this a lot and having a separate function for it simplifies the code, this would definitely be the place for it.

The function as written is incorrect though, since it should call `mock_block` to fill in all of the top-level properties, much like the `mmock_paragraph_block` does.

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,38 @@
+import logging
+
+from n2y.blocks import Block
+
+logger = logging.getLogger(__name__)
+
+class LinkToPage(Block):
```
johndgiese, line 7: The plugin class needs to inherit from the existing `n2y.blocks. LinkToPageBlock` class in order for this to work. (This could be made more explicit, but this is what's discussed in bullet point 2 in the README here https://github.com/innolitics/n2y#plugins )

johndgiese: I would have thought you'd get an exception raised from here because this is not subclassing `n2y.blocks.LinkToPageBlock`:

https://github.com/innolitics/n2y/blob/main/n2y/notion.py#L149-L155

Were you seeing this exception? If not, this is a bug in itself.

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,38 @@
...
+            self.client.get_page(self.linked_page_id) 
+            # Although out of the scope of the project at hand, get_page_or_database() 
+            # could be used in place of get_page, to cover the linked database case. 
+            # Might be an unnecessary step here. 
+            self.rich_text = self.client.wrap_notion_rich_text_array(self.notion_data["rich_text"], self)
+
+            # It's all about calling the proper method. 
+            content = self.rich_text.to_pandoc()
+
+        except PermissionError:
```
johndgiese, line 29: Nice job handling this exception case.

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,38 @@
...
+            msg = 'Permission denied when attempting to access linked page having id [%s]'
+            logger.warning(msg, self.notion_data)
```
johndgiese, line 31: I don't think `self.notion_data` is what you'd want to use here, since it will contain a lot more than the linked page id. Also, you'll want to be sure to include the URL to the LinkToPage block itself so that people who run into this error can quickly track down it's cause and fix it (by clicking on the URL and being taken to where the block is).

n2y/blocks.py
```diff
@@ -570,8 +570,17 @@ def to_pandoc(self):
...
 
 
-class LinkToPageBlock(WarningBlock):
-    pass
+class LinkToPageBlock(Block):
+    # Invoking warning, for backward compatibility with existing 
+    # installations, which will lack the plugin. 
+    # TODO: right? 
+
+    WarningBlock()
```
johndgiese, line 578: As best I can tell, you're instantiating a new instance of the `WarningBlock` class within the class definition block. The resulting instance is immediately discarded since it doesn't do anything, but even if it were assigned to something, I'm not sure what the intent would be.

@idolcemia could you expand a bit on what you're expecting this to do and why you'd think it would work?

n2y/test.py
```diff
@@ -0,0 +1,55 @@
+import logging
```
johndgiese, line 1: It looks like it's a test script you're using locally. I'd suggest not committing these sorts of files to the repo, as this sort of testing should really be handled in the automated tests.

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, client, notion_data, page, get_children=False):
+
+        super().__init__(client, page, notion_data, get_children)
+
+    def to_pandoc(self):
+        assert self.linked_page_id is not None
+
+        try:
+            self.client.get_page(self.linked_page_id) 
+            # Although out of the scope of the project at hand, get_page_or_database()
```
johndgiese, line 21: Good point about links to databases, I hadn't considered that! For now I would just log a warning if someone links to a database and then ignore it.

tests/test_blocks.py
```diff
@@ -255,6 +255,62 @@ def test_code_block():
     assert markdown == "``` javascript\nconst a = 3\n```\n"
 
 
+def test_link_to_page_block():
+
+# As I've been asked to update the end-to-end tests, but not specifically the unit tests
```
johndgiese, line 260: Ahh, if you'd like to add a unit test for the link too, that would be great!

tests/test_blocks.py
```diff
@@ -255,6 +255,62 @@ def test_code_block():
...
+
+    # Elaboration: As the link_to_page block is trivial, containing only the value of the link itself.
+    # And since representation of the page link in the pandoc output is not necessary --
+    # It is the contents of the linked page that is potentially
+    # represented, not the link to the page -- testing here would logically be limited to checking
+    # for proper conversion of the linked page.  However, such a test would be redundant, 
+    # as already handled in test-end-to-end.py
+
+    # Create the mock block of the linking page.
+    notion_block = mock_block("link_to_page", {
```
johndgiese, line 277: Nice job identifying the need to use `mock_block` here, but you're not calling it correctly. See the code for how mock_block is implemented to see why:

https://github.com/innolitics/n2y/blob/main/n2y/notion_mocks.py#L94

In particular, you can just use `mock_block("link_to_page", {"type": "page_id", "page_id":  "{mock_page_id}"})`

tests/test_end_to_end.py
```diff
@@ -256,6 +261,8 @@ def test_simple_page_to_markdown():
     assert "Page content" in document_as_markdown
 
 
+
+
 def test_builtin_plugins(tmp_path):
```
johndgiese, line 266: I was expecting that you would update the plugin end to end tests (i.e., this one where the comment is). See the issue for more details and let me know if it's not clear.

tests/test_end_to_end.py
```diff
@@ -212,8 +213,12 @@ def test_all_blocks_page_to_markdown(tmp_path):
     assert "``` javascript\nCode Block\n```" in document_as_markdown
     assert lines.count("This is a synced block.") == 2
     assert "This is a synced block from another page." in lines
+    assert "This page is not shared with the integration." in lines
+
     assert all(column_strings_in_lines) or (column_string in lines)
 
+
```
johndgiese, line 220: Flake8 will catch this, but these new lines are not pep8-compliant.

tests/test_blocks.py
```diff
@@ -10,7 +10,7 @@
     CodeBlock, BulletList, OrderedList, Decimal, Period, Meta, Pandoc, Link,
     HorizontalRule, BlockQuote, Image, MetaString, Table, TableHead, TableBody,
     TableFoot, RowHeadColumns, Row, Cell, RowSpan, ColSpan, ColWidthDefault, AlignDefault,
-    Caption, Math, DisplayMath
+    Caption, Math, DisplayMath,
```
johndgiese, line 13: I think flake8 will catch this, but you have some trailing whitespace here.

johndgiese (changes requested):
@idolcemia as you stated, I don't believe this is working. I left a number of comments which should help you identify why.

I realize this is a work in progress PR, but please note that most of the comments you've added should not be included in the final PR. Also, note that there are a lot of formatting issues that you'll need to fix in order to pass the flake8 check.

Let me know if you have any questions about my comments.

n2y/notion_mocks.py
idolcemia: Yes, I meant to clean that method out of there.  I realized some time ago that I didn't need a special mock.

n2y/plugins/linktopage.py
idolcemia: Yes, I do understand that the instructions were that the plugin should inherit from the LinkToPageBlock.  

I brought up some issues to Bimba concerning this inheritance.  I can elaborate.  Anyway, he suggested that I could just inherit from Block.  Sounded good to me, so as not to disturb the original functioning of the class for existing customers. I can explain to you.  The issue is described in my comments on files of previous commits. 

The only exception that I saw when trying to subclass n2y.blocks.LinkToPageBlock was that if the parent of that class were WarningBlock, then the interpreter complained that I wasn't overriding WarningBlock, as I recall.  I would need to recreate the error in order to tell you more.

n2y/plugins/linktopage.py
idolcemia: Thanks!

n2y/test.py
idolcemia: Absolutely.  Next time I will upload this as a separate file in another fashion  if I want you to see it.  I was unable to get the plugin loaded in the proper course of n2y, so I had to write this module to test it.  

test.py is **not** a test module, never mind the name.  It is a work around for my being unable to load as a plugin my plugin.

n2y/plugins/linktopage.py
johndgiese: I see, yeah @bimbashrestha @idolcemia as written, I believe you must inherit from LinkToPageBlock. I don't think you should need to modify the `n2y.blocks.LinkToPageClass` at all. Your class will presumably override both the `__init__` and `to_pandoc` methods, thus removing any warning functionality that you don't want in your class. Since you won't be touching `n2y.blocks.LinkToPageClass` at all, you can be sure it won't affect other people using it.

I'm surprised that it would complain about you not overriding WarningBlock. That sounds like a bug in the error handling code. If you happen to reproduce it, please copy the stack trace and let me know.

n2y/test.py
johndgiese: That makes sense! I think once you inherit from `n2y.blocks.LinkToPageBlock` that you'll be able to load your plugin.

tests/test_end_to_end.py
johndgiese: Per the issue, I was hoping you'd "Update the existing plugin end-to-end test". I.e., you would update the `test_builtin_plugins`. This end to end test already loads all of the other builtin plugins and tests that they work (more or less). You can just add your new plugin and update the associated notion page to test it.

I don't think you should need to add any new end to end tests (i.e.,  you can probably remove `test_links_to_pages`) or modify the `test_all_blocks_page_to_markdown` test (that test is meant to test how things work without any plugins.

tests/test_end_to_end.py
idolcemia: Yes, thanks.

tests/test_blocks.py
idolcemia: Got it.

idolcemia:
I've done everything, I think.  But I am still unable to test., for this reason:

When I look at page.blocks.children, the linked page block is represented as an unsupported block.  Since the blocks are not retrieved, they cannot be handled.  This condition predates my work. 

I may be missing something of course.

n2y/blocks.py
```diff
@@ -570,8 +570,30 @@ def to_pandoc(self):
...
+
+    def to_pandoc(self):
+        # TODO: in the future, if we are exporting the linked page too, then add
+        # a link to the page. For now, we just display the text of the page.
+
+        # The linked page may be outside the scope of what is accessible to n2y.
+        try:
+            page = self.client.get_page_or_database(self.linked_page_id)
+            title = page.title.to_pandoc()
+        except PermissionError:
```
johndgiese, line 590: Does `client.get_page_or_database` throw a `PermissionError` when the page is not accessible?

n2y/blocks.py
```diff
@@ -570,8 +570,30 @@ def to_pandoc(self):
...
+            msg = (
+                "Permission denied when attempting to access linked page having id [%s]"
+            )
+            logger.warning(msg, self.linked_page_id)
```
johndgiese, line 594: Please see the comment on my previous review here: https://github.com/innolitics/n2y/pull/69#discussion_r967007485

I don't think you fully addressed the comment.

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,58 @@
...
+        # TODO: Might be expanded to handle links to databases as well.
+        if self.link_type == "page_id":
+
+            try:
+                page = self.client.get_page_or_database(self.linked_page_id)
+
+                rich_text = self.client.wrap_notion_rich_text_array(
+                    page.notion_data["rich_text"], self)
+
+            except PermissionError:
```
johndgiese, line 35: See my other comment

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,58 @@
...
+    def to_pandoc(self):
+        assert self.linked_page_id is not None
+
+        # TODO: Might be expanded to handle links to databases as well.
+        if self.link_type == "page_id":
+
+            try:
+                page = self.client.get_page_or_database(self.linked_page_id)
+
+                rich_text = self.client.wrap_notion_rich_text_array(
```
johndgiese, line 32: @idolcemia can you explain the reasoning behind this line of code? I don't think it really makes sense and I'm surprised that the tests are passing.

n2y/plugins/linktopage.py
```diff
@@ -0,0 +1,58 @@
...
+                    "Permission denied when attempting to access linked page having id [%s]"
+                )
+                logger.warning(msg, self.linked_page_id)
+                return None
+
+            return rich_text.to_pandoc()
+
+        else:
+
+            logger.warning(
```
johndgiese, line 46: In general, all log messages should include URLs to the block in Notion where the error can be fixed. Thus, we should include a URL to the `LinkToPageBlock` where the link originates. The ids of the pages are less useful to a user. Thus, please update the error message to only include a link to the original block like most of the other `logger.warning` calls in the code repo. You can get a sense for what I mean by prepping for `logger.warning`.

@idolcemia I thought I'd spelled this out pretty clearly in the original Issue:

> Handle the case where the plugin doesn't have permission to access the linked page; in this case, log a warning including a link to the notion block where it occurred (grep for logger.warning for examples).

I also brought it up in my first review. Please pay attention to the details on this stuff.

tests/test_end_to_end.py
```diff
@@ -296,3 +338,15 @@ def test_builtin_plugins(tmp_path):
 def test_missing_object_exception():
     invalid_page_id = "11111111111111111111111111111111"
     assert run_n2y([invalid_page_id]) != 0
+
+
+def test_links_to_pages():
```
johndgiese, line 343: I thought, per the comments in the previous review, that we were going to remove `test_links_to_page`, especially given the comment you have here.

Here's my previous comment about this: https://github.com/innolitics/n2y/pull/69#discussion_r967368234

Please delete this test.

johndgiese (changes requested):
@idolcemia see my Slack comment for more details about the overall review. I don't think you've completed the last four items in the issue, so I unchecked them there. Once you push up the code that reproduces the error with the end-to-end test, then I can help debug the issue and then you'll be able to finish out the last four items.

johndgiese (approved):

PR #71, opened 2022-09-09
Fix a bug in the audit system

johndgiese:
The `Page.children` property was not recursing into child blocks and as a result any child database or page that was not a top-level block would be picked up by the audit system.

This bug would affect other things too, although it was noticed due to an audit producing false positives.

### DIFF
--- a/tests/test_audit_end_to_end.py
+++ b/tests/test_audit_end_to_end.py
@@ -5,7 +5,7 @@ from tests.utils import NOTION_ACCESS_TOKEN
 from n2y.audit import main
 
 
-def run_n2y(arguments):
+def run_n2yaudit(arguments):
     old_stdout = sys.stdout
     old_stderr = sys.stderr
     sys.stdout = StringIO()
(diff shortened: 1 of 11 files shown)


tests/test_audit_end_to_end.py
```diff
@@ -37,8 +37,11 @@ def test_audit():
...
     external_link_in_child_database = \
         'https://www.notion.so/B-4412005dcec24ff2827abbc367c90b29#6373a0b5c2804fbe9dfac167ce6948a0'  # noqa: E501
+    link_to_local_page_in_database_in_column = \
+        'https://www.notion.so/Audited-cfa8ff07bba244c8b967c9b6a7a954c1#21a13c06ef86462e882a181c6cb52a64'  # noqa: E501
 
     assert external_link_in_top_page in stdoutput
     assert external_link_in_child_page in stdoutput
     assert internal_link_in_child_page not in stdoutput
     assert external_link_in_child_database in stdoutput
+    assert link_to_local_page_in_database_in_column not in stdoutput
```
johndgiese, line 47: After adding this assertion, the test failed until I updated `Page.children`

sHermanGriffiths (approved):

Issue #54, opened 2022-08-10
Update to latest Notion API Version


johndgiese:

bimbashrestha:
The new Notion API requires making an API request for every property. This makes running n2y on pages or databases containing lots of properties impractical. @johndgiese contacted Notion support and they said they don't plan on deprecating older versions of the API, so we should be able to safely hold off on upgrading (for now).

Related WIP PR: https://github.com/innolitics/n2y/pull/64

PR #72, opened 2022-09-13
Upgrade Notion API to 2022-06-28

bimbashrestha:
The new API versions changes the following:
* ~Page properties must be retrieved using the page properties endpoint.~ (this no longer applies)
* Parents are now always direct parents; a parent field has been added to block. (this doesn't impact us)
* Database relations have a type of single_property and dual_property. (this doesn't impact us)

Tested it on one of our previous projects by running `make notion && make docs` before and after the API version change (with the same id_mappings.yml file) and then diffing the `data`, `documents`, and `img` directories. They were identical (`release` has diffs but that's not surprising since those consist docx and xlsx files)

Closes https://github.com/innolitics/n2y/issues/54

### DIFF
--- a/n2y/notion.py
+++ b/n2y/notion.py
@@ -85,7 +85,7 @@ class Client:
         self.headers = {
             "Authorization": f"Bearer {self.access_token}",
             "Content-Type": "application/json",
-            "Notion-Version": "2022-02-22",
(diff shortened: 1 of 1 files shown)


johndgiese (approved):
Nice work, that's great it was so easy!

PR #76, opened 2022-09-19
Persistent cache

idolcemia:
This is a work in progress.  I believe I've touched on all important mechanisms, but I have not yet run through it to double check functionality and some assumptions I made, and design decisions.  

In particular, I need to confirm for myself the data structure for child blocks, and what caching that block should do. 

I wrote the end to end test first, but haven't yet gone back to it.  It does not work as written.  

KNOWN ISSUE: This commit will not work and will fail all tests.  Please don't consider this finished work.


n2y/cache.py
```diff
@@ -0,0 +1,188 @@
...
+from os.path import exists
+
+from datetime import datetime
+
+logger = logging.getLogger(__name__)
+
+class Cache:
+    """Cache Notion block entities in local storage"""
+
+    def __init__(self, client):
```
johndgiese, line 14: I don't think we should pass `client` as an argument into `cache`, since the `cache` will actually be a child of the `client`. Thus, if we passed the client in it would be a sort of circular reference. Also, I don't think the cache should need to know about the client as best I'm aware.

n2y/cache.py
```diff
@@ -0,0 +1,188 @@
...
+        
+        self.client = client
+        self.cache_file = ".n2ycache"
```
johndgiese, line 17: We'll probably want to pass the full path of the cache file into the __init__ as an argument. Also, we'll want to create the file if it doesn't exist and initialize the SQL tables.

tests/test_cache.py
```diff
@@ -0,0 +1,13 @@
+from n2y.cache import Cache
+
+# Could also put these tests inside test_utils.py probably.
+
+
+def test_get_from_cache():
```
johndgiese, line 6: Nice, I would flesh these out before working any more on the real code.

tests/test_end_to_end.py
```diff
@@ -350,3 +350,53 @@ def test_builtin_plugins(tmp_path):
 def test_missing_object_exception():
     invalid_page_id = "11111111111111111111111111111111"
     assert run_n2y([invalid_page_id]) != 0
+
+
+def test_cache_effect(tmp_path):
```
johndgiese, line 355: Great! I would finish this test and make sure it's running correctly (but is failing) using a stub `Cache` class.

idolcemia:
The nature of the work at this point is all data, and we'd move this along very quickly if we had another set of eyes on it.  

Whom should I ask to help?

Just a matter of speed.  I think that someone like you could have it functionally done in an hour. 

Also, the issues of edited dates of pages and blocks is a topic that both you and Bimba are more up on than I.  I can do it for sure, but a little bit of help would make it go faster.

PR #78, opened 2022-09-20
Rework file naming system

sHermanGriffiths:


n2y/notion.py
```diff
@@ -366,30 +362,14 @@ def download_file(self, url, page):
...
-        with os.fdopen(temp_fd, 'wb') as temp_file:
-            for chunk in content_iterator:
-                hash_md5.update(chunk)
-                temp_file.write(chunk)
-
-        num_hash_characters = 8  # just long enough to avoid collisions
-        hash = hash_md5.hexdigest()[:num_hash_characters]
-        relative_filepath = "".join([page.filename, "-", hash, extension])
+    def save_file(self, content, page, extension):
+        page_id_chars = page.notion_id.replace('-', '')
```
johndgiese, line 369: Maybe use https://github.com/innolitics/n2y/blob/main/n2y/utils.py#L126 here? It's just one line, so maybe it's not really necessary to use the function.

n2y/notion.py
```diff
@@ -366,30 +362,14 @@ def download_file(self, url, page):
...
+        relative_filepath = "".join([page.filename, "-", page_id_chars[:11], extension])
```
johndgiese, line 370: I think either a format string or even just adding the strings together would be more idiomatic python here than using `str.join`

johndgiese (approved):
@sHermanGriffiths I had a few small suggested changes, but feel free to merge.

PR #81, opened 2022-09-23
Children

sHermanGriffiths:


tests/test_children.py
```diff
@@ -0,0 +1,25 @@
+from n2y.notion import Client
+from n2y.notion_mocks import mock_block, mock_rich_text
+from n2y.utils import id_from_share_link
+
+
+def test_create_and_delete_blocks():
+    url = "https://www.notion.so/innolitics/Test-Page-46bc615bed5a479e8390d9373279426f"
```
johndgiese, line 7: We should use the same notion pages and access token we use for the end to end tests here. See the doc string at the top of tests/end_to_end_tests.py

n2y/notion.py
```diff
@@ -373,3 +373,18 @@ def save_file(self, content, page, extension):
...
+    def append_block_children(self, block_id, children):
+        response = requests.patch(
+            f"{self.base_url}blocks/{block_id}/children",
+            json={"children": children}, headers=self.headers
+        )
+        return self._parse_response(response)
+
+    def delete_block(self, block_id):
+        headers = self.headers
+        del headers['Content-Type']
```
johndgiese, line 386: I think that deleting this here will delete it everywhere that else that uses `self.headers`, thus you may need to copy `self.headers` first or find some other way to remove Content-Type

tests/test_children.py
```diff
@@ -0,0 +1,25 @@
...
+            "rich_text": [mock_rich_text("m")],
+        },
+        has_children=True
+    )
+    del notion_block["url"]
+    creation_response = client.append_block_children(page.notion_id, [notion_block])
+    new_block_id = creation_response['results'][0]['id']
+    deletion_response = client.delete_block(new_block_id)
+    assert creation_response
+    assert deletion_response
```
johndgiese, line 25: This assertion seems to be just asserting that there is a `deletion_response`, but is that sufficient to make sure the deletion worked? It seems like even if it failed, `assert deletion_response` would pass just fine, although I may be missing something. To test this, try running `deletion_response = client.delete_block(new_block_id + "asdf")` and see if `assert deletion_response` fails. If it doesn't, that's a problem.

johndgiese (changes requested):
See my individual comments for details

n2y/notion_mocks.py
```diff
@@ -25,7 +25,7 @@ def mock_rich_text(text, annotations=None, href=None, mention=None):
         annotations = []
     if mention is None:
         rich_text_type = 'text'
-        content = {'content': text, 'link': None},
```
johndgiese, line 28: 👍

johndgiese (approved):
I had one small comment that needs to be addressed, but it looks good to merge otherwise.

PR #82, opened 2022-09-30
Config revamp

johndgiese:
@bimbashrestha this PR is ready for your review. I suggest reviewing commit-by-commit.

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -491,8 +491,11 @@ class WarningBlock(NoopBlock):
         return None
 
 
-class ChildDatabaseBlock(WarningBlock):
-    pass
+class ChildDatabaseBlock(NoopBlock):
+    def to_pandoc(self):
+        msg = 'Skipping unsupported "%s" block (%s). Perhaps you can convert the database into a simple table?'
+        logger.warning(msg, self.notion_type, self.notion_url)
+        return None
 
 
 class EmbedBlock(WarningBlock):
@@ -518,7 +521,7 @@ class VideoBlock(Block):
         return Para(content_ast)
 
 
-class PdfBlock(WarningBlock):
+class PdfBlock(Block):
     def __init__(self, client, notion_data, page, get_children=True):
         super().__init__(client, notion_data, page, get_children)
         self.pdf = client.wrap_notion_file(notion_data['pdf'])
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -493,7 +493,10 @@ class WarningBlock(NoopBlock):
 
 class ChildDatabaseBlock(NoopBlock):
     def to_pandoc(self):
-        msg = 'Skipping unsupported "%s" block (%s). Perhaps you can convert the database into a simple table?'
+        msg = (
+            'Skipping unsupported "%s" block (%s). '
+            'Perhaps you can convert the database into a simple table?'
+        )
         logger.warning(msg, self.notion_type, self.notion_url)
         return None
 
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -582,8 +582,13 @@ class LinkToPageBlock(Block):
     def to_pandoc(self):
         # TODO: in the future, if we are exporting the linked page too, then add
         # a link to the page. For now, we just display the text of the page.
+        if self.link_type == "page_id":
+            node = self.client.get_page(self.linked_page_id)
+        elif self.link_type == "database_id":
+            node = self.client.get_database(self.linked_page_id)
+        else:
+            raise NotImplementedError(f"Unknown link type: {self.link_type}")
 
-        node = self.client.get_page_or_database(self.linked_page_id)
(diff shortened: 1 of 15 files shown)


n2y/blocks.py
```diff
@@ -582,8 +582,13 @@ def __init__(self, client, notion_data, page, get_children=True):
     def to_pandoc(self):
         # TODO: in the future, if we are exporting the linked page too, then add
         # a link to the page. For now, we just display the text of the page.
+        if self.link_type == "page_id":
```
bimbashrestha, line 585: Is there a reason we're inlining this logic?

n2y/config.py
```diff
@@ -1,41 +1,134 @@
...
+    except yaml.YAMLError as exc:
+        logger.error("Error parsing the config file: %s", exc)
+        return None
+    except FileNotFoundError:
+        logger.error("The config file '%s' does not exist", path)
         return None
-    if not validate_database_config(config):
+    if not validate_config(config):
+        logger.error("Invalid config file: %s", path)
         return None
```
bimbashrestha, line 46: Maybe we can extract this try/except out?

```
def load_config(path):
    config = load_config_yaml(config_file)
    if config is None:
        return None
```

n2y/config.py
```diff
@@ -1,41 +1,134 @@
...
+    """
+    For each config item, merge in both the user provided defaults and the
+    builtin defaults for each key value pair."
+    """
+    merged_config_items = []
+    for config_item in config_items:
+        master_defaults_copy = copy.deepcopy(builtin_defaults)
+        defaults_copy = copy.deepcopy(defaults)
+        config_item_copy = copy.deepcopy(config_item)
+        merged_config_item = {**master_defaults_copy, **defaults_copy, **config_item_copy}
```
bimbashrestha, line 70: Didn't know you could update dictionaries this way. Cool!

n2y/config.py
```diff
@@ -1,41 +1,134 @@
...
+    # TODO: validate the export defaults key
+    return True
+
+
+def _validate_config_item(config_item):
+    if "id" not in config_item:
+        logger.error("Export config item missing the 'id' key")
+        return False
+    if not _valid_id(config_item["id"]):
+        logger.error("Invalid id in export config item: %s", config_item["id"])
```
bimbashrestha, line 94: Do we need a return False here?

n2y/config.py
bimbashrestha: I'm sure you looked into this but I assume that with this syntax, the dictionaries get updated in the order they appear right? So it's equivalent to:

```
merged_config_item = master_defaults_copy
merged_config_item = merged_config_item.update(defaults_copy)
merged_config_item = merged_config_item.update(config_item_copy)
```

n2y/main.py
```diff
@@ -103,130 +35,81 @@ def main(raw_args, access_token):
...
-    elif isinstance(node, Page):
-        print(node.to_markdown())
-    elif node is None:
-        msg = (
-            "Unable to find database or page with id %s. "
-            "Perhaps its not shared with the integration?"
-        )
-        logger.error(msg, object_id)
+    config = load_config(args.config)
+    if config is None:
```
bimbashrestha, line 47: Nice cleanup!

n2y/main.py
```diff
@@ -103,130 +35,81 @@ def main(raw_args, access_token):
...
-            if page.filename not in seen_file_names:
-                seen_file_names.add(page.filename)
-                with open(os.path.join(options.output, f"{page.filename}.md"), 'w') as f:
-                    f.write(page.to_markdown())
-            else:
-                logger.warning('Skipping page named "%s" since it has been used', page.filename)
-                counts['duplicate'] += 1
+    client = notion.Client(access_token, config["media_root"], config["media_url"])
+
+    for export in config['exports']:
```
bimbashrestha, line 52: Can we extract out the stuff inside this for loop for readibility?

Perhaps:

```
for export in config['exports']:
    client.load_plugins(export['plugins'])
    export_node_from_config(export)
```

tests/test_config.py
```diff
@@ -1,43 +1,119 @@
-from n2y.config import validate_database_config
+import copy
```
bimbashrestha, line 1: Nice tests!

bimbashrestha (approved):
Awesome work @johndgiese! Overall, it looks good to merge to me. Just some nit picks.

n2y/config.py
johndgiese: Yes! good catch!

n2y/blocks.py
johndgiese: I had thought that, since we know what type of link it is, it would be better to avoid the unnecessary API request that's present if we use `get_page_or_database`.

johndgiese:
FYI I didn't rebase because there were conflicts in the intermediate commits but not in the merge commit and it didn't seem worth sorting them out.

PR #83, opened 2022-10-03
From pandoc

sHermanGriffiths:
- [x] Check that you've followed the Dependencies Procedure for any new dependencies.
- [x] ...


n2y/blocks.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
johndgiese, line 1: What's this for?

n2y/blocks.py
```diff
@@ -80,6 +81,16 @@ def children_to_pandoc(self):
                         pandoc_ast.append(result)
         return pandoc_ast
 
+    def _retrieve_chidren(self):
```
johndgiese, line 84: Since we added this, perhaps we could simplify the `__init__` method?

n2y/notion.py
```diff
@@ -160,14 +162,45 @@ def get_class_list(self, notion_object, object_type=None):
         except KeyError:
             raise NotImplementedError(f'Unknown "{notion_object}" class of type "{object_type}"')
 
-    def instantiate_class(self, notion_object, object_type, *args, **kwargs):
+    def instantiate_class(self, *args, **kwargs):
+        first_arg = args[0]
+        if type(first_arg) == str:
```
bimbashrestha, line 167: If we want to use `instantiate_class` for both pandoc and notion, then we should probably make it clearer what `first_arg` is. Perhaps even explicitly pull it out of `*args` like:

```
def instantiate_class(self, payload, *args, **kwargs):
    instantiating_notion = payload == str
    if instantiating_notion:
        ...
```
or even more explicit, add a new argument.

```
def instantiate_class(self, instantiating_notion, *args, **kwargs):
    if instantiating_notion:
        ...
```

tests/test_from_pandoc.py
```diff
@@ -0,0 +1,30 @@
+from n2y.notion import Client
+from n2y.utils import id_from_share_link
+
+
+url = "a36972712dd24d808bb0107bf0945f41"
+token = "secret_aBbNlu5xsTtBNarAmuPYliS6pSCMpNaIJkcsz5eFuZn"
+client = Client(token, plugins=None)
+object_id = id_from_share_link(url)
+page = client.get_page_or_database(object_id)
```
bimbashrestha, line 9: I like that you're extracting out code that needs to be run before every test. 

The problem with having these in the global scope like this is that the tests are no longer independent. For example, you're re-using the same `client` in each of your tests so if one test alters the state of your `client`, another test might be affected inadvertently by that. This can lead to strange or flaky test failures.

Instead, I'd recommend either wrapping all the non-literal variables (client, page) in a helper method or better yet, using pytest's fixtures: https://docs.pytest.org/en/6.2.x/fixture.html. They're designed for exactly this use case. 

Hope that makes sense.

n2y/notion.py
bimbashrestha: BTW, I'm not opposed to just having two separate methods `instantiate_class_from_notion` and `instantiate_class_from_pandoc` and not having `instantiate_class` at all. Or perhaps leaving `instantiate_class` as is and just adding `instantiate_class_from_pandoc`. 

I think I could be convinced either way but this seems like another approach worth considering.

bimbashrestha (changes requested):
Nice start! Haven't looked into this in great detail but left a few comments.

Also please fix the flake8 complaints from the ci :). It blocks the actual tests from getting run.

n2y/blocks.py
sHermanGriffiths: accidental import

n2y/blocks.py
sHermanGriffiths: I added that unnecessarily, I've just deleted it

n2y/notion.py
sHermanGriffiths: I like the idea of explicitly pulling the first argument from args. I've assigned it the name `class_to_instantiate`, as the argument will either literally be a pandoc class to instantiate or a string denoting which class to instantiate. Instantiate class now looks like this:
```
def instantiate_class(self, class_to_instantiate, *args, **kwargs):
        if type(class_to_instantiate) == str:
            return self._instantiate_class_from_notion(class_to_instantiate, args[0], *args[1:], **kwargs)
        else:
            return self._instantiate_class_from_pandoc(class_to_instantiate, args[0], **kwargs)
```

tests/test_from_pandoc.py
sHermanGriffiths: pytest fixtures are really cool; thanks for the tidbit. Definitely the perfect place to employ them

tests/test_from_pandoc.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+
+from n2y.notion import Client
+from n2y.utils import id_from_share_link
+
+
+@pytest.fixture
+def client():
+    token = "secret_aBbNlu5xsTtBNarAmuPYliS6pSCMpNaIJkcsz5eFuZn"
```
johndgiese, line 9: Why not use https://github.com/innolitics/n2y/blob/main/tests/utils.py#L14 ?

tests/test_blocks.py
```diff
@@ -6,37 +6,10 @@
...
-    Row,
-    Cell,
-    RowSpan,
-    ColSpan,
-    ColWidthDefault,
-    AlignDefault,
-    Caption,
-    Math,
-    DisplayMath,
+    Str, Para, Plain, Header, CodeBlock, BulletList, OrderedList, Decimal,
```
johndgiese, line 9: 👍

n2y/errors.py
```diff
@@ -26,6 +26,14 @@ class UseNextClass(Exception):
     pass
 
 
+class UseNextType(Exception):
```
johndgiese, line 29: I was thinking about this; what's the benefit of creating a new exception class here, instead of using the existing `UseNextClass` exception?

n2y/notion.py
```diff
@@ -92,13 +93,23 @@ def get_default_classes(self):
...
+        return pandoc_types
+
     def load_plugins(self, plugins):
         self.notion_classes = self.get_default_classes()
         if plugins is not None:
             for plugin in plugins:
                 plugin_module = importlib.import_module(plugin)
                 try:
                     self.load_plugin(plugin_module.notion_classes)
+                    if "notion_classes" in plugin_module.__dict__:
```
johndgiese, line 111: What's the purpose of this line?

n2y/errors.py
sHermanGriffiths: It just created it for specificity, you'd know there was an error in instantiating a class from a pandoc type if this error was thrown and an error in instantiating a class from notion if the other was. It's deprecated now, as I access the classes directly.

n2y/notion.py
sHermanGriffiths: sorry, legacy

PR #84, opened 2022-10-04
Emoji support

johndgiese:


n2y/main.py
```diff
@@ -27,11 +28,6 @@ def main(raw_args, access_token):
         help="Level to set the root logging module to",
     )
 
-    # TODO: Add the ability to dump out a "schema" file that contains the schema
```
johndgiese, line 30: I removed these since they seem kind of out of scope and a long way off, if we did ever want to add them.

n2y/blocks.py
```diff
@@ -491,8 +491,14 @@ def to_pandoc(self):
         return None
 
 
-class ChildDatabaseBlock(WarningBlock):
-    pass
+class ChildDatabaseBlock(NoopBlock):
```
bimbashrestha, line 494: Nice!

n2y/emoji.py
```diff
@@ -0,0 +1,14 @@
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class Emoji:
```
bimbashrestha, line 7: I love this haha

bimbashrestha (approved):
If it won't take long, maybe we can add an emoji to our end to end tests?

Feel free to merge without though.

johndgiese:
@bimbashrestha I looked into testing this, but it would take some effort to test mainly because the end-to-end tests don't have access to the `Page` objects and the `page.icon` isn't used at all in the exports. Thus, to test the emoji we would need to create an end-to-end test that worked lower level or we would need to create a plugin to extract out the emoji. I do have one page in the end-to-end tests that has an emoji, which ensures it won't crash if there is an emoji :)

I'm going to merge.

PR #86, opened 2022-10-14
Copy to

sHermanGriffiths:
# Author's Checklist

- [x] **Automated Checks:** Confirm that the linter and automated tests are passing.
- [x] **Verification:** Atomic unit tests have been added, if appropriate.
- [x] **Verification:** End-to-end tests have been extended or updated, if appropriate.
- [x] **Low-level Documentation:** Comments or doc strings for affected code has been updated.
- [x] **High-Level Documentation:** The README is updated as appropriate.
- [ ] **Change Log:** New features or bug fixes have been added to the README.
Not sure I should be changing the README for these changes. If so, what should I add?

# Content for the Reviewer

- [ ] Should be reviewed commit-by-commit
- [ ] Should be reviewed all at once


johndgiese (changes requested):
Unless this is a draft PR, please complete the checklist instructions prior to submitting the PR for review.

sHermanGriffiths:
> Unless this is a draft PR, please complete the checklist instructions prior to submitting the PR for review.

@johndgiese sorry, just finished. Forgot you were notified upon request.

n2y/blocks.py
```diff
@@ -48,16 +49,17 @@ def __init__(self, client, notion_data, page=None, get_children=True):
...
-                children = []
+        if get_children and self.has_children:
+            children = self.client.get_child_blocks(self.notion_id, page, get_children)
         else:
-            children = None
+            children = Children(client=self.client)
         self.children = children
 
+    def replicate(self):
+        return self.client.wrap_notion_block(self.notion_block, self.page, self.get_children)
```
johndgiese, line 61: When we replicate a block, perhaps we should remove the UUID from `self.notion_block` first, that way we don't have duplicate UUIDs anywhere?

n2y/blocks.py
```diff
@@ -604,6 +606,389 @@ def to_pandoc(self):
...
+        elif '_block' in self.destination.__dict__:
+            self.page = self.destination
+        else:
+            self.page = self.destination.page
+        if not self.page:
+            raise NotImplementedError(
+                'Page must be defined.'
+            )
+
+    def _default_info(self):
```
johndgiese, line 977: All of this should really go in `__init__` so that people reading the class can see where all of the properties get defined.

n2y/blocks.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, client, notion_data, page=None, get_children=True):
         self.client = client
         self.page = page
 
+        self.notion_block = notion_data
```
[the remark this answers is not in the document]
sHermanGriffiths: `self.notion_block = notion_data` whereas `self.notion_data = notion_data[notion_data['type']]`.  I think renaming what is currently self.notion_data to self.notion_block and vice versa would more accurate to what the api calls them but they do have different values. I wanted to start saving the original dictionary that comes in because depending on the block type, sometimes information is lost or mutated when parsing it from key-values to class attributes. Thus, instead of having to reconstruct the class instantiation's `notion_block` form by creating a dictionary or mutating `block.__dict__` (which is sometimes impossible due to the lossy process I mentioned earlier), I figured we can just save it when it comes in.

n2y/blocks.py
sHermanGriffiths: true, I deleted this method though as it's now dead code.

n2y/blocks.py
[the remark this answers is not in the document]
johndgiese: @sHermanGriffiths my main concern is that, although they have different values, `notion_data[notion_data['type']]` can be derived from just `notion_data`. Thus, storing both of them is redundant. I hadn't thought we needed all of the content of `notion_data`, which is why previously we used `notion_data[notion_data['type']]`. However, now that we do need everything in `notion_data`, I think we can just always save everything.

n2y/notion.py
```diff
@@ -313,6 +314,11 @@ def get_notion_page_property(self, page_id, property_id):
         response = requests.get(url, headers=self.headers)
         return self._parse_response(response)
 
+    def create_notion_page(self, page_data):
```
johndgiese, line 317: 👍

n2y/notion.py
```diff
@@ -375,13 +381,67 @@ def save_file(self, content, page, extension):
...
+            },
+        }
+        response = requests.post(
+            f"{self.base_url}comments",
+            headers=self.headers,
+            json=data
         )
         return self._parse_response(response)
 
+    def create_notion_database(self, notion_data):
```
johndgiese, line 438: 👍

n2y/notion_mocks.py
```diff
@@ -20,15 +20,15 @@ def mock_rich_text_array(text_blocks_descriptors):
     if isinstance(text_blocks_descriptors, str):
         return [mock_rich_text(text_blocks_descriptors, [])]
     else:
-        return [mock_rich_text(t, a) for t, a in text_blocks_descriptors]
```
johndgiese, line 23: 👍

n2y/notion.py
```diff
@@ -375,13 +381,67 @@ def save_file(self, content, page, extension):
             temp_file.write(content)
         return urljoin(self.media_url, relative_filepath)
 
-    def append_block_children(self, block_id, children):
-        response = requests.patch(
-            f"{self.base_url}blocks/{block_id}/children",
-            json={"children": children}, headers=self.headers
+    def append_block_children(self, block_id, children, page_or_database_info=None):
```
johndgiese, line 384: Per our call:

Rewrite to remove the need for `page_or_database_info`.

If child pages work, keep that.

If child database don't work, remove the code for this and raise NotImplementedError

johndgiese: Rename to `append_notion_blocks`

johndgiese: Add a note to the doctoring saying that not all blocks allow you to add children.

n2y/blocks.py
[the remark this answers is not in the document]
sHermanGriffiths: sorry I misunderstood.  Would you still be okay with saving the value of `notion_data[notion_data['type']] ` as `block.notion_block`? I only ask because otherwise, we'd be replacing the use of `block.notion_data` with `block.notion_data[block.notion_type]`. I still think it would be worthwhile to save such a wordy and widely used value as a class attribute.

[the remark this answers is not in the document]
johndgiese: It just feels bad to have duplicate copies of the data. How about this:

```
@property
def notion_type_data(self):
  return self.notion_data[self.notion_data["type"]]
```

This avoids duplicating the data or the code for pulling out the Notion type data. Note you'll still need to refactor the naming a bit.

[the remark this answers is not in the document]
sHermanGriffiths: Perfect, thank you!

n2y/notion.py
sHermanGriffiths: great news, appending both child_databases and child_pages works perfectly with this new method!

johndgiese (approved):
Nice work Sam’an, this was a tricky issue but I think we iterated towards a great solution!

PR #92, opened 2022-12-08
Fix bug by naming files by the block id instead of the page id

sHermanGriffiths:
# Author's Checklist

  N/A
  N/A
  N/A
- [x] **High-Level Documentation:** The README is updated, as appropriate.


n2y/plugins/mermaid.py
```diff
@@ -64,8 +65,14 @@ def to_pandoc(self):
                 '-o', temp_filepath,
             ], capture_output=True, input=diagram_as_bytes, check=True)
             with open(temp_filepath, 'rb') as temp_file:
-                content = temp_file.read(4096)
-                url = self.client.save_file(content, self.page, '.png')
+                content = temp_file.read()
+                root = root = Path(__file__).resolve().parent.parent
```
bimbashrestha, line 69: Hm is this a typo? `root = root =`?

tests/utils.py
```diff
@@ -13,7 +13,6 @@
 # will need to be used to create them.
 NOTION_ACCESS_TOKEN = 'secret_lylx4iL5awveY3re6opuvSQqM6sMRu572TowhfzPy5r'
 
-
```
bimbashrestha, line 16: NIt

n2y/plugins/mermaid.py
```diff
@@ -64,8 +65,14 @@ def to_pandoc(self):
...
+                with open(root/'data'/'mermaid_err.png', 'rb') as err_img:
+                    if content == err_img.read():
```
bimbashrestha, line 71: Hey @sHermanGriffiths I'm confused as to what's happening here. Can you explain it a bit? Is this error image something the mermaid cli creates by default when there is a syntax error and we're checking to see if the bytes match to see if an error was thrown? 

I haven't looked into this but I'd definitely see if there is another way to handle this since checking for an entire image seems a bit odd.

bimbashrestha (changes requested):
Some comments. Good work!

n2y/plugins/mermaid.py
sHermanGriffiths: That's exactly right. Since mermaid doesn't throw an error, the only way to know there was an error is if this image is rendered.

PR #94, opened 2022-12-19
Add features that allow the notion Client to retry api calls after be…

sHermanGriffiths:
…ing rate limited

# Author's Checklist

  N\A
  N\A
  N\A
  N\A

### DIFF
--- a/n2y/main.py
+++ b/n2y/main.py
@@ -5,9 +5,9 @@ import argparse
 import pkg_resources
 
 from n2y import notion
-from n2y.export import export_page, database_to_yaml, database_to_markdown_files
 from n2y.config import load_config
-from n2y.utils import share_link_from_id
+from n2y.utils import share_link_from_id, DEFAULT_MAX_RETRIES
+from n2y.export import export_page, database_to_yaml, database_to_markdown_files
 
 logger = None
 
@@ -25,7 +25,7 @@ def main(raw_args, access_token):
     )
     parser.add_argument("config", help="The path to the config file")
     parser.add_argument(
-        '--max_retries', '-m', default=5,
+        '--max_retries', '-m', default=DEFAULT_MAX_RETRIES,
         help=(
             'The maximum amount of times an API request will be retried after being rate '
             'limited. This argument must be an integer greater than or equal to 5'
--- a/n2y/main.py
+++ b/n2y/main.py
@@ -25,6 +25,13 @@ def main(raw_args, access_token):
     )
     parser.add_argument("config", help="The path to the config file")
     parser.add_argument(
+        '--max_retries', '-m', default=5,
+        help=(
+            'The maximum amount of times an API request will be retried after being rate '
+            'limited. This argument must be an integer greater than or equal to 5'
+        )
+    )
+    parser.add_argument(
         "--verbosity", '-v', default='INFO',
         help="Level to set the root logging module to",
     )
@@ -48,7 +55,12 @@ def main(raw_args, access_token):
     if config is None:
         return 2
 
-    client = notion.Client(access_token, config["media_root"], config["media_url"])
+    client = notion.Client(
+        access_token,
+        config["media_root"],
+        config["media_url"],
+        max_retries=args.max_retries
+    )
 
     error_occurred = False
     for export in config['exports']:
--- a/n2y/notion.py
+++ b/n2y/notion.py
@@ -57,7 +57,9 @@ def retry_api_call(api_call):
                 'retry_api_call must be a n2y.notion `Client` object'
             )
         try:
-            return api_call(*args, **kwargs)
+            response = api_call(*args, **kwargs)
+            client.retry_count = 0
+            return response
         except APIResponseError as err:
             can_retry = 'retry-after' in err.headers
             if can_retry and client.retry_api_calls:
--- a/n2y/notion.py
+++ b/n2y/notion.py
@@ -3,6 +3,7 @@ import logging
 import requests
 import functools
 import importlib.util
+from datetime import datetime, timedelta
 from time import sleep
 from os import path, makedirs
 from urllib.parse import urljoin, urlparse
@@ -64,12 +65,12 @@ def retry_api_call(api_call):
             can_retry = 'retry-after' in err.headers
             if can_retry and client.retry_api_calls:
                 client.retry_count += 1
-                retry_after = err.headers['retry-after']
+                retry_after = float(err.headers['retry-after'])
                 logger.info(
                     'Client has been rate limited. This API call '
                     f'will be retried in {retry_after} seconds'
                 )
-                sleep(int(retry_after))
+                sleep(retry_after)
                 return wrapper(*args, **kwargs)
             else:
                 client.retry_api_calls or logger.warning(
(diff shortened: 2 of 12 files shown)


n2y/notion.py
```diff
@@ -84,6 +116,21 @@ def __init__(
         self.load_plugins(plugins)
         self.plugin_data = {}
 
+    @property
+    def retry_api_calls(self):
+        return self.max_retries > self.retry_count
+    
+    @property
+    def max_retries(self):
```
johndgiese, line 124: @bimbashrestha and @sHermanGriffiths I was thinking the "max retries" should apply to a single API call. Thus, if an API call gets a 429, it will retry that same API call up to 5 times.

If the next API call after that gets a 429, it could also get up to 5 retries before an exception is raised. Is that the behavior you guys were expecting?

sHermanGriffiths: Yes, I forgot to reset the count. Just finished doing that

n2y/main.py
```diff
@@ -24,6 +24,13 @@ def main(raw_args, access_token):
         formatter_class=argparse.RawTextHelpFormatter,
     )
     parser.add_argument("config", help="The path to the config file")
+    parser.add_argument(
+        '--max_retries', '-m', default=5,
```
bimbashrestha, line 28: We have the default set to 5 in multiple places. Let's extract the 5 to a constant somewhere and use that everywhere. That way, we can just change it in one place when we need to.

n2y/notion.py
```diff
@@ -46,6 +47,35 @@
 # TODO: Rename this file `client.py`
 
 
+def retry_api_call(api_call):
```
bimbashrestha, line 50: I like that you're using a decorator. Nice idea!

I know there are a few popular python libraries that provide a retry decorator ([retry](https://github.com/invl/retry), [tenacity](https://github.com/jd/tenacity) are two that I know of). I think we should use one of them instead of implementing our own. You can probably swap their decorator for yours as-is. It'll also make it easier for us to add more advanced retry functionality like backoffs, jitter, choose which exceptions to retry on, etc later on. What do you think @sHermanGriffiths and @johndgiese?

Also @sHermanGriffiths, if we do implement our own decorator, we need a good unit test to make sure it works.

n2y/notion.py
```diff
@@ -84,6 +117,21 @@ def __init__(
...
+    def retry_api_calls(self):
+        return self.max_retries > self.retry_count
+
+    @property
+    def max_retries(self):
+        return self._max_retries
+
+    @max_retries.setter
+    def max_retries(self, val):
+        if int(val) > 5:
```
bimbashrestha, line 130: Extract constant max retries.

n2y/notion.py
```diff
@@ -66,10 +96,13 @@ def __init__(
         media_root='.',
         media_url='',
         plugins=None,
+        max_retries=5,
```
bimbashrestha, line 99: Extract constant max retries

n2y/rich_text.py
```diff
@@ -24,10 +24,9 @@ class RichText:
     a property_value or somewhere else.
     """
 
-    def __init__(self, client, notion_data, block=None, table=False):
+    def __init__(self, client, notion_data, block=None):
```
bimbashrestha, line 27: Is this change related to retries?

tests/test_plain_text_to_pandoc.py
```diff
@@ -21,8 +21,3 @@ def test_plain_text_to_pandoc_spaces_after():
 def test_plain_text_to_pandoc_spaces_after_newline():
     pandoc_ast = RichText.plain_text_to_pandoc("hello\n  world")
     assert pandoc_ast == [Str("hello"), LineBreak(), Space(), Space(), Str("world")]
-
-
-def test_plain_text_to_pandoc_table():
```
bimbashrestha, line 26: Why are we deleting this test case?

bimbashrestha (changes requested):
Nice start @sHermanGriffiths! Requested some changes.

Also, I noticed that this PR contains changes for more than just retries. Let's move those to their own PR off of `main` so that it can be reviewed on it's own. 

Combining multiple things into a PR makes sense if they are all small items or if they are related to each other. The retry stuff seems pretty independent from the table stuff. If they're not, let me know.

n2y/notion.py
bimbashrestha: Nice catch. Yeah, each api call should definitely be treated independently in terms of retries.

n2y/notion.py
johndgiese: I like the idea of using a library for this.

sHermanGriffiths: I like the idea of using a library as well, but I can't find any that parse the delay time from the exception itself, which is the premise of the functionality we're trying to create here.

johndgiese: @sHermanGriffiths that's a great point, I think we should use the header sent back from Notion to figure out the timeout time.

bimbashrestha: @sHermanGriffiths Ah I see. Great point and that makes total sense!

I'll leave the decision up to you. If you do move forward with a custom implementation though, please include a test for it.

tests/test_utils.py
```diff
@@ -43,3 +45,69 @@ def test_page_id_from_share_link():
...
+        elif client.retry_count == 3:
+            assert isclose(14, seconds, abs_tol=1)
+            raise APIResponseError(Resp(2), '', 439)
+    client = Client('', max_retries=3)
+    try:
+        tester(client, datetime.now())
+    except APIResponseError:
+        assert True
+    else:
+        assert False
```
bimbashrestha, line 113:
```suggestion
    with pytest.raises(APIResponseError):
        tester(client, datetime.now())
```

n2y/utils.py
```diff
@@ -34,10 +37,6 @@ def processed_date_to_plain_text(processed_date):
         return processed_date
 
 
-# see https://pandoc.org/MANUAL.html#exit-codes
-PANDOC_PARSE_ERROR = 64
```
bimbashrestha, line 38: Is deleting this intentional?

bimbashrestha (changes requested):
@sHermanGriffiths Nice work! 

This is close. I left some inline comments (mostly small). 

Let's also add a test for making sure retry counts are independent for each api call. The thing David mentioned. You can just call trigger your decorator two times in a row and make sure the retry count is reset for the second run.

tests/test_utils.py
```diff
@@ -43,3 +45,69 @@ def test_page_id_from_share_link():
...
     assert id_from_share_link(share_link) == page_id
+
+
+def test_retry_api_call_no_error():
+    @retry_api_call
+    def tester(client):
+        assert client.max_retries == 5
+        assert client.retry_count == 0
+        assert client.retry_api_calls
+        return 5
```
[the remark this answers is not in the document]
sHermanGriffiths: Yes, I've changed it to just return true

n2y/utils.py
sHermanGriffiths: not at all

bimbashrestha (approved):
Nice work!

bimbashrestha (changes requested):
Ah actually looks like there are some CI failures, so I'm re-requesting changes

bimbashrestha (approved):

PR #95, opened 2022-12-21
rework table richtext bug fix

sHermanGriffiths:
# Author's Checklist

  N\A
  N\A
  N\A
  N\A
  N\A

### DIFF
--- a/tests/test_plain_text_to_pandoc.py
+++ b/tests/test_plain_text_to_pandoc.py
@@ -21,8 +21,3 @@ def test_plain_text_to_pandoc_spaces_after():
 def test_plain_text_to_pandoc_spaces_after_newline():
     pandoc_ast = RichText.plain_text_to_pandoc("hello\n  world")
     assert pandoc_ast == [Str("hello"), LineBreak(), Space(), Space(), Str("world")]
-
-
-def test_plain_text_to_pandoc_table():
-    pandoc_ast = RichText.plain_text_to_pandoc("hello\n  world", True)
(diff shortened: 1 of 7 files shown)


tests/test_plain_text_to_pandoc.py
```diff
@@ -21,8 +21,3 @@ def test_plain_text_to_pandoc_spaces_after():
 def test_plain_text_to_pandoc_spaces_after_newline():
     pandoc_ast = RichText.plain_text_to_pandoc("hello\n  world")
     assert pandoc_ast == [Str("hello"), LineBreak(), Space(), Space(), Str("world")]
-
-
-def test_plain_text_to_pandoc_table():
```
bimbashrestha, line 26: I assume you're deleting this test case because it breaks after your changes? Let's fix the test case so that it works with your changes.

bimbashrestha (changes requested):
Nice work @sHermanGriffiths! 

Add a quick test for your change and this looks good to go. Maybe we can just edit the test case you deleted?

bimbashrestha:
Thanks for pulling this out into it's own PR! Reviewing it like this is so much easier :)

tests/test_plain_text_to_pandoc.py
sHermanGriffiths: I created it as a test case for the previous changes

bimbashrestha (approved):
@sHermanGriffiths please add a test that catches the bug you fixed in this PR before merging. Good job with the fix.

PR #96, opened 2023-01-21
Rdm

sHermanGriffiths:
# Author's Checklist

  N/A
  N/A
  N/A
  N/A


n2y/blocks.py
```diff
@@ -63,7 +63,6 @@ def to_pandoc(self):
         raise NotImplementedError()
 
     def children_to_pandoc(self):
-        assert self.has_children
```
bimbashrestha, line 66: Is this intentional? If so, why are we removing it?

n2y/plugins/render.py
```diff
@@ -0,0 +1,210 @@
...
+            template_name = f'{id}-template.md'
+            open(output_name, 'x').close()
+            with open(template_name, 'x') as file:
+                file.write(pandoc.write(ast))
+            render_template_to_file(config, template_name, context, output_name)
+            with open(output_name, 'r') as file:
+                rendered_ast = pandoc.read(file.read())
+            return rendered_ast[1][0]
+        finally:
+            os.remove(output_name)
```
bimbashrestha, line 202: Can we use the python `tempfile` module to create/delete these?

tests/test_end_to_end.py
```diff
@@ -312,3 +312,69 @@ def test_comment():
     comments = Client(NOTION_ACCESS_TOKEN).get_comments(block_with_comments_id)
     assert comments[0].rich_text.to_plain_text() == "Test Comment"
     assert comments[1].rich_text.to_plain_text() == "Test Comment 2"
+
+
+def test_render_plugin(tmpdir):
+    def _run_n2y(temp_dir, config, render_config):
```
bimbashrestha, line 318: Don't we already have this method in this file? Can we re-use  that instead of creating a new one?

bimbashrestha (changes requested):
Nice start @sHermanGriffiths. 

I tried to use your changes to export DHF-templates and wasn't able to. Here is what I tried:

* Reordered the exports inside `clients/template/notion-qms.yml` so that the data files were pulled first. 
* Added n2y.plugins.render as a a plugin for all the `database_as_files` exports.
* `make notion`

I expected there to be no jinja in the `documents` folder after this but there were. 

Can you have a go? I might've done something incorrectly btw. 

You've got a pretty good end-to-end test. Nice work with that! But I think the real test will be getting these changes to work on the DHF-template workspace.

n2y/blocks.py
sHermanGriffiths: yes, because if it's called on a block with no children, we want it to return an empty list instead of throwing an error.

sHermanGriffiths:
> Nice start @sHermanGriffiths.
> 
> I tried to use your changes to export DHF-templates and wasn't able to. Here is what I tried:
> 
> * Reordered the exports inside `clients/template/notion-qms.yml` so that the data files were pulled first.
> * Added n2y.plugins.render as a a plugin for all the `database_as_files` exports.
> * `make notion`
> 
> I expected there to be no jinja in the `documents` folder after this but there were.
> 
> Can you have a go? I might've done something incorrectly btw.
> 
> You've got a pretty good end-to-end test. Nice work with that! But I think the real test will be getting these changes to work on the DHF-template workspace.

 I tested that as well and it does work. Did you change the makefile? It needs a `--render-config` argument on the n2y calls.

n2y/plugins/render.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+from n2y.blocks import FencedCodeBlock
+from n2y.errors import UseNextClass
+
+
+logger = logging.getLogger(__name__)
+
+
+class FirstPassOutput:
```
johndgiese, line 20: Be sure to pull over all RDM tests related to this code.

n2y/notion.py
```diff
@@ -116,6 +118,7 @@ def __init__(
 
         self.databases_cache = {}
         self.pages_cache = {}
+        self.yaml_cache = {}
```
johndgiese, line 121: I think we should be able to avoid the need for another cache, since we can instead rely on the databases_cache.

bimbashrestha:
> I tested that as well and it does work. Did you change the makefile? It needs a --render-config argument on the n2y calls.

@sHermanGriffiths that's great!

Hmm, I believe I included the --render-config. Do you mind pushing up your changes to the DHF-template `Makefile` and `notion_qms.yml` to a branch on the `notion-qms` repo so I can try it out locally?

n2y/plugins/render.py
sHermanGriffiths: yes

tests/test_end_to_end.py
sHermanGriffiths: this way lets me customize the config file

bimbashrestha (changes requested):
Discussed offline. Requesting changes to put this back on your queue.

bimbashrestha (changes requested):

bimbashrestha (approved):

PR #100, opened 2023-01-27
expand childpages if converted to pandoc

sHermanGriffiths:
# Author's Checklist

  N/A
  N/A
  N/A
  N/A
  N/A

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -70,8 +70,6 @@ class Block:
             elif block_type == ChildPageBlock:
                 for b in blocks:
                     result = b.to_pandoc()
-                    title = result[0][0]['title'][0]
-                    pandoc_ast.append(Header(1, ('', [], []), [Str(title)]))
                     pandoc_ast.extend(result[1])
             else:
                 for b in blocks:
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -67,14 +67,12 @@ class Block:
         for block_type, blocks in groupby(self.children, lambda c: type(c)):
             if issubclass(block_type, ListItemBlock):
                 pandoc_ast.append(block_type.list_to_pandoc(blocks))
-            elif block_type == ChildPageBlock:
-                for b in blocks:
-                    result = b.to_pandoc()
-                    pandoc_ast.extend(result[1])
             else:
                 for b in blocks:
                     result = b.to_pandoc()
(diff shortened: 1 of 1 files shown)


n2y/blocks.py
```diff
@@ -67,6 +67,12 @@ def children_to_pandoc(self):
         for block_type, blocks in groupby(self.children, lambda c: type(c)):
             if issubclass(block_type, ListItemBlock):
                 pandoc_ast.append(block_type.list_to_pandoc(blocks))
+            elif block_type == ChildPageBlock:
+                for b in blocks:
+                    result = b.to_pandoc()
+                    title = result[0][0]['title'][0]
```
johndgiese, line 73: Nice idea! I think it makes a lot of sense to expand out sub pages. I would actually not show the page title as an H1, and would just show the page's content in place. This would work identically to linked pages then, which we use (or have a plugin) that expands.

sHermanGriffiths: yea, I only did it as a debug because [this](https://www.notion.so/Type-is-used-to-categorize-requirements-Possible-types-include-cde96c39f96141a9b43f165d422addbf) subpage was converting to pandoc as a Pandoc block inside of another Pandoc block and that's against protocol. Is there another way you'd solve that?

johndgiese: I think this was due to a weird thing Emily did in the templates. I just fixed it in the templates! That said, I think expanding the page (without the title) is a good idea.

bimbashrestha (changes requested):

n2y/blocks.py
```diff
@@ -67,6 +67,10 @@ def children_to_pandoc(self):
         for block_type, blocks in groupby(self.children, lambda c: type(c)):
             if issubclass(block_type, ListItemBlock):
                 pandoc_ast.append(block_type.list_to_pandoc(blocks))
+            elif block_type == ChildPageBlock:
```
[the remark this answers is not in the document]
sHermanGriffiths: It's not the same because ChildPageBlock.to_pandoc() results in a Pandoc block which will cause an error because all of the pandoc will eventually end up in another Pandoc block. that's why `result[1]` is selected on line 73. I could bring this logic down into that iteration though

bimbashrestha (approved):
Nice work @sHermanGriffiths!

PR #113, opened 2023-04-04
specify columns pandoc option

sHermanGriffiths:
# Author's Checklist

  N/A
  N/A
  N/A
  N/A
  N/A

### DIFF
--- a/n2y/config.py
+++ b/n2y/config.py
@@ -24,6 +24,7 @@ EXPORT_DEFAULTS = {
     "pandoc_format": "markdown",
     "pandoc_options": [
         '--wrap', 'none',  # don't hard line-wrap
+        '--columns', '10000',  # don't cap the number of columns in tables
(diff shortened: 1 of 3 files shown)


johndgiese (approved):

n2y/config.py
```diff
@@ -24,6 +24,7 @@
     "pandoc_format": "markdown",
     "pandoc_options": [
         '--wrap', 'none',  # don't hard line-wrap
+        '--columns', '10000',  # don't cap the number of columns in tables
```
johndgiese, line 27: @sHermanGriffiths I think this could use a more detailed comment here. Perhaps something like:

The default column width is 72 characters. When the width is this small, then pandoc may elect to generate HTML tables in markdown instead of text-based tables; this is problematic when we then convert the markdown into DOCX files which don't support raw HTML.

PR #115, opened 2023-04-11
Filter

sHermanGriffiths:
# Author's Checklist

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -158,7 +158,10 @@ class BulletedListItemBlock(ListItemBlock):
         content = [Plain(self.rich_text.to_pandoc())]
         if self.has_children:
             children = self.children_to_pandoc()
-            content.extend(children)
+            for child in children:
+                if isinstance(child, Table):
+                    content.append(Para([]))
+                content.append(child)
         return content
 
     @classmethod
@@ -186,7 +189,10 @@ class NumberedListItemBlock(ListItemBlock):
         content = [Plain(self.rich_text.to_pandoc())]
         if self.has_children:
             children = self.children_to_pandoc()
-            content.extend(children)
+            for child in children:
+                if isinstance(child, Table):
+                    content.append(Para([]))
+                content.append(child)
         return content
 
     @classmethod
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -160,6 +160,9 @@ class BulletedListItemBlock(ListItemBlock):
             children = self.children_to_pandoc()
             for child in children:
                 if isinstance(child, Table):
+                    # A bug in pandoc's markdown reader makes it unable to register mid-list tables.
+                    # This adds an extra space between the table and the previous list item, which
+                    # Allows the table to be read as such.
                     content.append(Para([]))
                 content.append(child)
         return content
@@ -191,6 +194,7 @@ class NumberedListItemBlock(ListItemBlock):
             children = self.children_to_pandoc()
             for child in children:
                 if isinstance(child, Table):
+                    # See comment in BulletedListItemBlock.to_pandoc()
                     content.append(Para([]))
                 content.append(child)
         return content
--- a/n2y/notion.py
+++ b/n2y/notion.py
@@ -184,11 +184,11 @@ class Client:
         """
         page_in_cache = notion_data["id"] in self.pages_cache
         if page_in_cache and self.page_class_is_in_use(
-        # Need to check that the page in the client cache was instantiated using
-        # The currently favored page class. Otherwise, plugins set for one export
-        # Will be used in another or a plugin will be set but not used.
-        # As we currently use the `jinjarenderpage` plugin for all pages,
-        # this check is most likely uneccessary at this point.
+            # Need to check that the page in the client cache was instantiated using
+            # the currently favored page class. Otherwise, plugins set for one export
+            # will be used in another or a plugin will be set but not used.
+            # As we currently use the `jinjarenderpage` plugin for all pages,
+            # this check is most likely uneccessary at this point.
             self.pages_cache[notion_data["id"]]
         ):
             return self.pages_cache[notion_data["id"]]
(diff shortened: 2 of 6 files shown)


johndgiese (approved):
Nice work!

n2y/plugins/jinjarenderpage.py
```diff
@@ -120,8 +153,10 @@ def _create_jinja_environment():
         extensions=["jinja2.ext.do"],
     )
     environment.globals['first_pass_output'] = FirstPassOutput()
-    environment.filters['join_to'] = join_to
+    environment.filters['fuzzy_find_in'] = fuzzy_find_in
+    environment.filters['fuzzy_search'] = fuzzy_search
```
johndgiese, line 157: I'd delete "fuzzy_search" and "fuzzy_in" if we're not using them.

n2y/plugins/jinjarenderpage.py
```diff
@@ -142,19 +177,25 @@ def render_from_string(source, context=None, environment=None):
...
         super().__init__(client, notion_data)
-        self.client.plugin_data['jinjarenderpage'] = {
-            'environment': _create_jinja_environment(),
-        }
+        if 'jinjarenderpage' not in client.plugin_data:
+            client.plugin_data['jinjarenderpage'] = {}
+        if self.notion_id not in client.plugin_data:
+            client.plugin_data['jinjarenderpage'][self.notion_id] = {
+                'environment': _create_jinja_environment(),
+                'database_cache': {},
```
johndgiese, line 185: Let's remove the "database_cache" here. We already cache at the network request level (see https://github.com/innolitics/n2y#cache).

https://wiki.c2.com/?PrematureOptimization=

n2y/plugins/jinjarenderpage.py
```diff
@@ -142,19 +177,25 @@ def render_from_string(source, context=None, environment=None):
 class JinjaRenderPage(Page):
     def __init__(self, client, notion_data):
         super().__init__(client, notion_data)
-        self.client.plugin_data['jinjarenderpage'] = {
-            'environment': _create_jinja_environment(),
```
johndgiese, line 146: Add a comment to these lines https://github.com/innolitics/n2y/blob/main/n2y/notion.py#L185-L192

explaining the problem that lead to us checking the page plugin class type. Then note that now we typically use the same plugins for all pages, and thus the check is likely no longer necessary.

I'd also add a logger warning call noting that the existing page class got replaced.

n2y/blocks.py
```diff
@@ -158,7 +158,10 @@ def to_pandoc(self):
         content = [Plain(self.rich_text.to_pandoc())]
         if self.has_children:
             children = self.children_to_pandoc()
-            content.extend(children)
```
johndgiese, line 161: Try wrapping in a paragraph.

Also, add a detailed comment explaining why we needed to do this.

n2y/blocks.py
```diff
@@ -186,7 +189,10 @@ def to_pandoc(self):
...
```
johndgiese, line 189: Add comment pointing to the other comment

n2y/blocks.py
sHermanGriffiths: Paragraph wrap didn't work

n2y/notion.py
```diff
@@ -182,12 +182,22 @@ def _wrap_notion_page(self, notion_data):
         replace our existing page instance, along with it's content or other
         state that has been added to it.
         """
-        if notion_data["id"] in self.pages_cache and self.page_class_is_in_use(
+        page_in_cache = notion_data["id"] in self.pages_cache
+        if page_in_cache and self.page_class_is_in_use(
+        # Need to check that the page in the client cache was instantiated using
+        # The currently favored page class. Otherwise, plugins set for one export
```
johndgiese, line 188:
```suggestion
        # the currently favored page class. Otherwise, plugins set for one export
```

n2y/notion.py
```diff
@@ -182,12 +182,22 @@ def _wrap_notion_page(self, notion_data):
...
+        # Will be used in another or a plugin will be set but not used.
```
johndgiese, line 189:
```suggestion
        # will be used in another or a plugin will be set but not used.
```

n2y/plugins/jinjarenderpage.py
```diff
@@ -190,21 +220,32 @@ def __init__(self, client, notion_data, page, get_children=True):
...
         context = {
             "databases": self.databases,
             "page": self.page.properties_to_values(self.pandoc_format),
         }
+        if 'render_content' in jinja_code:
+            def render_content(notion_id):
+                page = self.client.get_page(notion_id)
+                for child in page.block.children:
+                    if isinstance(child, HeadingBlock):
+                        child.level += 1
```
johndgiese, line 236: Instead of hard-coding the header indent at `+= 1` here, I think we should add an optional second argument to the `render_content` filter that defaults to 0, but can be set to any positive or negative number. If a negative number is used, we need to be careful to ensure the resulting header level is larger than 1. Thus, if the header level is 3 and a header adjustment of -3 is passed in, then the result should still be 1. Perhaps:

```
child.level = max(1, child.level + level_adjustment)
```

johndgiese (approved):
@sHermanGriffiths I had a few small suggested changes. Also, the automated checks need to pass. Otherwise, this looks good!

n2y/plugins/jinjarenderpage.py
sHermanGriffiths: I think it would make more sense to default it to 1 so that the default functionality is to maintain consistency with the page headers as a whole but otherwise I agree.

johndgiese: I think the default should be to keep the content as-is, since I wouldn't expect a filter like this to automatically indent and it could surprise people if it mutates the content automatically. If we kept the +1, then we'd probably want to change the name of the filter to indicate this behavior, but then the name may be too long. (That said, having this option is a great idea.)

PR #119, opened 2023-05-12
What to do about empty headers necessary to satisfyMarkdown table spec

matt-hayden:
Tests the potential for tables to be rendered with blank header rows. The requirements for this Issue changed from a bugfix to a reaction to what the formats can support.

I left some opinions as my own comments, not in scope for this.

# Author's Checklist (TODO)

- [n/a] **Requirements:** Review the [requirements](https://www.notion.so/medtech-os/Issue-with-Simple-Tables-that-have-a-Header-Column-9020ff64287f4bf6870ab53adb3eb378?pvs=4) in the issue for this PR and confirm they're all implemented.


tests/end_to_end/test_tables.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    object_id = "9b1dd705f61647b6a10032ec7671402f"
+
+    config = {
+        "exports": [
+            {
+                "id": object_id,
+                "node_type": "page",
+                "output": f"{request.node.name}-output",
+                "pandoc_format": "gfm",
```
matt-hayden, line 33: This format is key. gfm=GitHub-flavored Markdown. We might get better use from other dialects of Markdown.

n2y/page.py
```diff
@@ -101,7 +103,22 @@ def parent(self):
             return self.client.get_database(self.notion_parent["database_id"])
 
     def to_pandoc(self):
-        return self.block.to_pandoc()
```
johndgiese, line 104: @matt-hayden ideally this would not go in the page-level block, but rather the table level block. Is there a reason it needs to be a the page level? Ideally we'll keep the page level pretty thin if we can.

matt-hayden: No, and if we don't need to know how many tables are affected in total, then this can all live in Table methods.

n2y/export.py
```diff
@@ -58,6 +59,17 @@ def export_page(
 ):
     page_properties = _page_properties(page, pandoc_format, id_property, url_property, property_map)
     pandoc_ast = page.to_pandoc()
+
+    n_empty_headers = 0
```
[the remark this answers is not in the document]
matt-hayden: We talked about the problem being a little beyond the class organization, so I reintroduced it back into the export method. Being closer to export means we might tune the logic based on knowing the destination format.

tests/end_to_end/test_tables.py
```diff
@@ -0,0 +1,83 @@
+import pytest  # noqa: E401
```
matt-hayden: I _think_ this is considered best-practice nowadays; pytest modules should begin with import pytest, even if unused.

tests/end_to_end/test_tables.py
```diff
@@ -0,0 +1,83 @@
...
+        ]
+    }
+    with monkeypatch.context() as m:
+        m.chdir(tmp_path)
+        config_path = Path("{}-config.yaml".format(request.node.name))
+        with config_path.open("w") as fo:
+            yaml.dump(config, fo)
+        m.setattr(sys, "argv", ["n2y", str(config_path)])
+        # cli_main calls sys.exit, which should be changed for testing
+        m.setattr(sys, "exit", lambda x: x)
```
matt-hayden: We might consider ditching the cli_main wrappers because of the smell of patching sys.exit.

matt-hayden (commented):
I would pursue:
- Refactor `__main__` scoped calls so that main() takes no arguments and does not call sys.exit
- Prefix the package-level CLI with commands "export" and "audit"

n2y/export.py
```diff
@@ -58,6 +59,23 @@ def export_page(
...
+            if isinstance(element, Table):
+                _, head = element[3]
+                n_header_rows = sum(1 for _, row in head)
+                if n_header_rows == 0:
+                    n_empty_headers += 1
+        if n_empty_headers:
+            logger.warning(
+                "%d table(s) will present empty headers to maintain Markdown spec",
+                n_empty_headers,
+                extra={"url": page.notion_url}
```
johndgiese, line 76: Let's print out the URL directly (or does this already do this?). See other log messages to see how this is normally done.

n2y/export.py
```diff
@@ -58,6 +59,23 @@ def export_page(
 ):
     page_properties = _page_properties(page, pandoc_format, id_property, url_property, property_map)
     pandoc_ast = page.to_pandoc()
+    has_tables = pandoc_ast and any(isinstance(e, Table) for e in pandoc_ast[1])
+
+    if has_tables:
```
johndgiese, line 64: Perhaps you could split this out into a separate function, say "_num_headerless_tables"?

n2y/export.py
```diff
@@ -58,6 +59,14 @@ def export_page(
 ):
     page_properties = _page_properties(page, pandoc_format, id_property, url_property, property_map)
     pandoc_ast = page.to_pandoc()
+
+    if (n_empty_headers := _count_headerless_tables(pandoc_ast)) > 0:
```
johndgiese, line 63: I propose that we split this into two lines to make it simpler

```
num_empty_headers = _count_headerless_tables(pandoc_ast)
if num_empty_headers > 0:
 ....
```

johndgiese: I think the rest of the codebase uses "num_" instead of "n_". I'd see what's used elsewhere in the codebase and follow that.

johndgiese (changes requested):
See other comments

tests/end_to_end/test_tables.py
```diff
@@ -0,0 +1,99 @@
...
+        "exports": [
+            {
+                "id": object_id,
+                "node_type": "page",
+                "output": f"{request.node.name}-output",
+                "pandoc_format": "gfm",
+            }
+        ]
+    }
+    with monkeypatch.context() as m:
```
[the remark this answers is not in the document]
matt-hayden: I think there are a few lessons that've made pytest more opinionated about this. People trying to run concurrently (test or plugin) will hit thread problems when their process calls an os.chdir or sets an environment variable, so the advice is to monkeypatch and/or use absolute paths. There are other examples, too, and I'm not debating their relevance but whether they come officially from within that project.

n2y/export.py
```diff
@@ -58,6 +59,14 @@ def export_page(
...
+        logger.warning(
+            "%r: %d table(s) will present empty headers to maintain Markdown spec",
```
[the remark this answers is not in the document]
matt-hayden: OK so for plugins it looks like square brackets, and core code in parens. I brought all core URLs into the paren-at-the-end syntax.

tests/end_to_end/test_tables.py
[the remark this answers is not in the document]
johndgiese: I can see how calling `os.chdir` wouldn't work in a multi-threaded environment. I don't think the approach in https://github.com/innolitics/n2y/blob/main/tests/test_end_to_end.py#L18-L46 uses environment variables.

If this is easy to fix and the solution is simple and doesn't make the code much more complicated, I suggest fixing it here:

https://github.com/innolitics/n2y/blob/main/tests/test_end_to_end.py#L18-L46

and making your test follow the approach of the existing tests.

If it is not easy to fix, please create an issue in the task backlog and explain the problem and a proposed solution and approximately how much time you think it will take to fix. We can then decide if it's important to fix during our next sprint planning call.

Overall, the goals are to (1) focus on keeping the code simple and following existing patterns that are already set, (2) meet the requirements of the specified tasks in a timely manner.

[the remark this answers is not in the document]
matt-hayden: It probably doesn't rise to be an Issue, and my experience is that fitting pytest fixtures into helper functions does add boilerplate -- would be counterproductive to goal (1). To meet goal (2) I just refactored this into the test_end_to_end module. Should be much shorter amount to review.

n2y/blocks.py
```diff
@@ -627,7 +627,7 @@ def to_pandoc(self):
             raise NotImplementedError(f"Unknown link type: {self.link_type}")
 
         if node is None:
-            msg = "Permission denied when attempting to access linked node [%s]"
+            msg = "Permission denied when attempting to access linked node (%r)"
```
johndgiese, line 630: 👍 nice!

johndgiese (approved):
Nice work!

PR #122, opened 2023-05-22
New footnote plugin

dwhensley:
Added new `dbfootnotes` plugin. Major features:
- Supports footnotes as page mentions where each footnote is a page in a footnotes database (and which will typically be an inline database).
- To be promoted to a footnote, a page mention must come from a database with a title ending with "Footnotes".
- Confirmed footnotes are implemented with the `Note` type in the Pandoc AST for general export (i.e., not specialized for markdown only).

# Author's Checklist

- [x] **Low-level Documentation:** Comments or doc strings for affected code have been updated.
- [x] **High-Level Documentation:** The README is updated, as appropriate.

### DIFF
--- a/n2y/plugins/dbfootnotes.py
+++ b/n2y/plugins/dbfootnotes.py
@@ -14,17 +14,26 @@ logger = logging.getLogger(__name__)
 class PageMentionFootnote(PageMention):
     def __init__(self, client, notion_data, plain_text, block=None):
         super().__init__(client, notion_data, plain_text, block)
-        self.mentioned_page = self.client.get_page(self.notion_page_id)
-        if self._is_footnote():
+        if (
+            self._get_mention_page()
+            and self._get_mention_page_parent()
+            and self._is_footnote()
+        ):
             self._attach_footnote()
         else:
             raise UseNextClass
 
+    def _get_mention_page(self):
+        self.mentioned_page = self.client.get_page(self.notion_page_id)
+        return self.mentioned_page is not None
+
+    def _get_mention_page_parent(self):
+        self.mentioned_page_parent = self.mentioned_page.parent
+        return self.mentioned_page_parent is not None
+
     def _is_footnote(self):
-        if self.mentioned_page is not None and isinstance(
-            self.mentioned_page.parent, Database
-        ):
-            return self.mentioned_page.parent.title.to_plain_text().endswith(
+        if isinstance(self.mentioned_page_parent, Database):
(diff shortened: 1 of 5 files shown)


n2y/plugins/dbfootnotes.py
```diff
@@ -0,0 +1,50 @@
...
+from n2y.errors import UseNextClass
+from n2y.mentions import PageMention
+
+plugin_data_key = "n2y.plugins.dbfootnotes"
+
+logger = logging.getLogger(__name__)
+
+
+class PageMentionFootnote(PageMention):
+    def __init__(self, client, notion_data, plain_text, block=None):
```
johndgiese, line 15: I'd probably raise `UseNextClass` if `block` is None... this may happen implicitly already since perhaps `get_page` will fail...

johndgiese (approved):
@dwhensley the code looks great! Simple and easy to follow, and I like the tests.

Do we check if someone accidentally mentions a footnotes database on a different page? If not, I think we should add this and should raise an exception if this happens. We should also log an error message that includes a URL to where the mention originates. You'll probably need to go from the mention to the containing block (if there is one---sometimes rich text exists in properties and other places where it's not in a block).

Once this side case is addressed feel free to merge!

n2y/plugins/dbfootnotes.py
dwhensley: :+1: Fixed.

dwhensley:
> @dwhensley the code looks great! Simple and easy to follow, and I like the tests.
> 
> Do we check if someone accidentally mentions a footnotes database on a different page? If not, I think we should add this and should raise an exception if this happens. We should also log an error message that includes a URL to where the mention originates. You'll probably need to go from the mention to the containing block (if there is one---sometimes rich text exists in properties and other places where it's not in a block).
> 
> Once this side case is addressed feel free to merge!

@johndgiese 
We were not checking for DB refs on a different page, good callout. This has been fixed to now emit a `PluginError` Exception with URL information as you suggested. This added a bit more logic and updates to the tests. I also edited the `flake8` ignore list because I noticed `W503` and `W504`, which contradict each other, were both active. This is described in [this closed PR](https://github.com/mila-iqia/cookiecutter-pyml/issues/10). This led to an impossible situation where a chain of binary comparisons needs to be split across lines. I ignored both of them for now, but if you think we should keep to one style, we can add that back in.

Given all of these changes, I've requested a re-review. Thanks!

johndgiese:
@dwhensley these changes look good! Please merge at your earliest convenience!

PR #121, opened 2023-05-19
Support embedding metadata into docx

matt-hayden:
This is in draft one of the requirements strictly implies a pandoc types introspection not carried into the export method.

https://github.com/innolitics/notion-qms/pull/61 stacks on this.

# Author's Checklist

Before assigning reviewers, step through the following checklist. Check each item once it's completed. If an item is skipped, write an explanation below the item.

- [ ] **Construction:** Review the code diff and confirm there's no dead code, debug code, or unnecessary comments.
- [ ] **Requirements:** Review the requirements in the issue for this PR and confirm they're all implemented.
- [ ] **Architecture:** Confirm that the changes to the code are consistent with the existing architecture.
- [ ] **SOUP/OTS Software:** Any new dependencies have appropriate licenses and are documented or pinned in requirements files.
- [ ] **Unit Test:** Atomic unit tests have been added, if appropriate.
- [ ] **End-to-End Tests:** End-to-end tests have been extended or updated, if appropriate.
- [ ] **Low-level Documentation:** Comments or doc strings for affected code have been updated.
- [ ] **High-Level Documentation:** The README is updated, as appropriate.
- [ ] **Change Log:** New features or bug fixes have been added to the change log in the README.

PR #124, opened 2023-05-26
Support metadata in formats other than Markdown

matt-hayden:
Rather than whittle the YAML front matter by hand, this burdens Pandoc to saddlebag the metadata into formats that support it. 

Side-effect: apparently Pandoc believes that an empty input deserves no metadata. A skim of the pandoc source makes me believe it exits early, which means this PR needs to introduce a workaround which inserts a dummy value into empty documents. 

This is a cherry pick around the force merge to other-pandoc-formats. It should replace https://github.com/innolitics/n2y/pull/121, before the force merge.

# Author's Checklist

- [see note] **Requirements:** Review the requirements in the issue for this PR and confirm they're all implemented. The header-footer script https://github.com/innolitics/notion-qms/pull/61 would depend on this PR, and contains the requirements.

- [ ] **Architecture:** Confirm that the changes to the code are consistent with the existing architecture.
- [see note] **End-to-End Tests:** Tests for Markdown pass, but DOCX has a different bug that carries only some of the metadata to Pandoc, which we have not tested for. 


n2y/export.py
```diff
@@ -66,20 +69,24 @@ def export_page(
...
-            yaml.dump(page_properties) + '---',
-            page_content,
-        ])
-    else:
-        # if the result is a binary file, return it as is (since we can't add YAML metadata to it)
-        return page_content
+    if not pandoc_ast:
+        logger.warning("Page is empty, inserting a placeholder to satisfy Pandoc (%r)",
+                       page.notion_url)
+        pandoc_ast = [Para([Str("𐞁")])]
```
matt-hayden, line 75: This is a squiggly

tests/test_end_to_end.py
```diff
@@ -149,11 +159,16 @@ def test_simple_database_to_docx_files(tmpdir):
...
     assert status == 0
-    output_directory = os.path.join(tmpdir, "database")
-    generated_files = {f for f in listdir(output_directory) if isfile(join(output_directory, f))}
-    assert generated_files == {"A.docx", "B.docx", "C.docx"}
+    for expected_filename in ["A.docx", "B.docx", "C.docx"]:
+        expected_path = tmp_path / "database" / expected_filename
+        doc = pandoc.read(expected_path.read_bytes(), format="docx")
+        metadata_dict = doc[0][0]
+
+        # Should arrive from yaml
```
matt-hayden, line 169: This is where I expected a comment -- does each record in a simple database inherit these values? I assumed so, and maybe it's harmless.

PR #140, opened 2023-08-08
ignore casing of terms matched by the fuzzy_find_in jinja filter

sHermanGriffiths:
# Describe Your Changes
Modified the `re.finditr` call to `re.IGNORECASE`  so that lower cased references to the key terms still get matched

# How Did You Test It
I exported SW-0010_MIC_Finder_SDDS_Rev_C from Specific-Dx and ensured that Software Item was included in the glossary, as it's referenced in the document, although in lowercase. I ran the debugger on my n2y code to see if the terms were coming through correctly and they were so that indicated to me that the matching system was filtering out the desired terms. I immediately knew it was the casing so I added the aforementioned flag and that fixed the issue

Screen Shot 2023-08-08 at 3 03 19 PM
Screen Shot 2023-08-08 at 3 01 13 PM


bimbashrestha (approved):
Awesome! Nice test write up

PR #141, opened 2023-08-11
convert custom class to parent class for yaml dump

sHermanGriffiths:
# Describe Your Changes
Convert properties into dictionaries before yaml dump

# How Did You Test It
I tested this by exporting a database as yaml until it formatted correctly. I tried modifying the custom class in different ways to make it dump like a dictionary but I guess you can't do that for dictionary-like custom classes. Actually, I FOUND A WAY TO DO ITT!!!! Had to kind of deep dive the pyyaml docs but it works! I found a way but it doesn't work with the testing module so I reverted back to just converting it to a dict in our functions instead of inside pyyaml.

### DIFF
--- a/n2y/export.py
+++ b/n2y/export.py
@@ -11,6 +11,7 @@ from n2y.logger import logger
 from n2y.utils import (
     pandoc_format_to_file_extension,
     pandoc_write_or_log_errors,
+    custom_representer,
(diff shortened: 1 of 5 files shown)


bimbashrestha (approved):
Looks good. Thanks for writing the test plan!

PR #142, opened 2023-08-29
Toc

sHermanGriffiths:
# Describe Your Changes
I built and implemented a table of contents block and a table of contents item block. I also modified the codebase to use common syntax practices wherever possible and updated the version of Python used to build the workflow release.

# How Did You Test It
I built a config file (see Figure 1) for a page with a table of contents notion block, then exported it and tweaked the code in succession until it worked as expected (see Figure 2). I then wrote tests for the `TableOfContentsItemBlock` and `TableOfContentsBlock` classes [here](https://github.com/innolitics/n2y/blob/a2393200174971880e5eac1c3bc826eccef82312/tests/test_blocks.py#L684C1-L724C4) so that if they are changed, we can identify which is causing the problem more effectively.

## Figure 1:
Screen Shot 2023-08-31 at 1 07 25 PM

## Figure 2:
Screen Shot 2023-08-31 at 1 08 24 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.8
+      - name: Set up Python 3.11.4
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.11.4
       - name: Install pypa/build
         run: python -m pip install build --user
       - name: Build a binary wheel and a source tarball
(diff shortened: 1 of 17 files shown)


.github/workflows/release.yml
```diff
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.8
+      - name: Set up Python 3.11.4
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.11.4
```
bimbashrestha, line 15: 👍🏽

PR #145, opened 2023-09-08
Use conventions wherever possible and update workflow python versions

sHermanGriffiths:
# Describe Your Changes
- Update all workflows to use Python 3.11.4
- Create variables for repeatedly created objects
- Add default values where conventional
- Use the conventional `cls` variable instead of `klass` in class method creation

# How Did You Test It
I made sure that all tests in our testing suite were still passing.


bimbashrestha (approved):
Looks good. I assume this PR introduces no functional change and that it's just a refactor right? Feel free to merge

PR #146, opened 2023-09-08
Build and implement TableOfContentsBlock class

sHermanGriffiths:
# Describe Your Changes
I built and implemented a table of contents block and a table of contents item block.

# How Did You Test It
I built a config file (see Figure 1) for a page with a table of contents notion block, then exported it and tweaked the code in succession until it worked as expected (see Figure 2). I then wrote tests for the `TableOfContentsItemBlock` and `TableOfContentsBlock` classes [here](https://github.com/innolitics/n2y/pull/146/files#diff-c35ecd7f598a9ea24ee0b03107fa01b029252319eb42e12baf6c438bbe22dcebR685-R724) so that if they are changed, we can identify which is causing the problem more effectively.

## Figure 1:
Screen Shot 2023-08-31 at 1 07 25 PM

## Figure 2:
Screen Shot 2023-08-31 at 1 08 24 PM

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -9,7 +9,8 @@ from pandoc.types import (
 )
 
 from n2y.logger import logger
-from n2y.utils import yaml_map_to_meta, header_id_from_text
+from n2y.notion_mocks import mock_block, mock_rich_text_array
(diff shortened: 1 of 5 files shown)


bimbashrestha (approved):
Looks good to me!

PR #155, opened 2023-09-29
Handle header trees starting with a non-H1 header 

russellkan:
# Describe Your Changes
- Avoid `IndexError` when header trees start with a non-H1 header
- Skip out of order headers with a warning
- Add unit test cases

# How Did You Test It
-  Ran `pytest tests` with all tests passing

PR #156, opened 2023-09-30
Fix bugs causing our publish-documents workflow to fail

russellkan:
# Context
After merging #155, I manually ran the `publish-documents` workflow, which failed with a different error: https://github.com/innolitics/innolitics.github.io/actions/runs/6357254369/job/17268129193

This was due to a `BulletList` being passed [here](https://github.com/innolitics/n2y/blob/8ac2aa97afe7fbf38b1a4eeb654157887de0caae/n2y/blocks.py#L683C8-L684).

To fix this, we first check if the `ast` is an array, and if so, run `filter_linebreaks` on that.

After fixing this, a `KeyError` error occurred in a local run [here](https://github.com/innolitics/n2y/blob/8ac2aa97afe7fbf38b1a4eeb654157887de0caae/n2y/database.py#L85).

~With the fix, it just logs a warning and returns `None`:~

The `KeyError` was caused by the parent page being a database, so the `page_id` key did not exist. The fix checks the `type` (which should be one of: `workspace`, `block_id`, or `page_id`) and returns the appropriate parent (`None` for `workspace`, `database` for `block_id`, and `page` for `page_id`.

# How Did You Test It
- Ran `pytest tests` with all tests passing
- Installed locally and run `n2y publish_config.yml --verbosity=INFO` in the site repo successfully.

### DIFF
--- a/n2y/database.py
+++ b/n2y/database.py
@@ -81,8 +81,15 @@ class Database:
     def parent(self):
         if self.notion_parent["type"] == "workspace":
             return None
-        else:
-            return self.client.get_page(self.notion_parent["page_id"])
+
+        if "page_id" not in self.notion_parent:
+            logger.warning(
+                f"Parent of database {self.title.to_plain_text()} (id: {self.notion_id})"
+                f" has no page_id: {self.notion_parent.items()}"
+            )
+            return None
+
+        return self.client.get_page(self.notion_parent["page_id"])
 
     def to_pandoc(self):
         return self.block.to_pandoc()
--- a/n2y/database.py
+++ b/n2y/database.py
@@ -79,17 +79,22 @@ class Database:
 
     @property
     def parent(self):
-        if self.notion_parent["type"] == "workspace":
+        """
+        The parent of a database can be a page, another database, a block, or
+        it can be at the top level of workspace.
+        """
+        type = self.notion_parent["type"]
+        if type == "workspace":
             return None
 
-        if "page_id" not in self.notion_parent:
-            logger.warning(
-                f"Parent of database {self.title.to_plain_text()} (id: {self.notion_id})"
-                f" has no page_id: {self.notion_parent.items()}"
-            )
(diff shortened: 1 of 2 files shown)


n2y/database.py
```diff
@@ -81,8 +81,15 @@ def _init_icon(self, icon_notion_data):
     def parent(self):
         if self.notion_parent["type"] == "workspace":
             return None
-        else:
-            return self.client.get_page(self.notion_parent["page_id"])
+
+        if "page_id" not in self.notion_parent:
```
johndgiese, line 85: When would this happen? E.g., is it because the parent is actually a database? I think the only valid parent types are workspace, database, and page. Instead of returning none here, perhaps there's a better way to handle the case?

johndgiese (changes requested):

n2y/database.py
russellkan: Ah, that makes sense. The new commit returns the database by using the `block_id`.

n2y/database.py
```diff
@@ -79,9 +79,17 @@ def _init_icon(self, icon_notion_data):
...
+        """
+        The parent should be a workspace, database ("block_id"), or page ("page_id").
+        """
+        type = self.notion_parent["type"]
+        if type == "workspace":
             return None
-        else:
+
+        if type == "block_id":
+            return self.client.get_database(self.notion_parent["block_id"])
```
johndgiese, line 90: @russellkan are you sure this works? I'm surprised that "block_id" would be a valid database_id here ...

I think this is incorrect. See https://developers.notion.com/reference/parent-object

russellkan: You're right; it's not a valid database_id but it did pass without error. I took a look at the API reference and updated it so it returns the appropriate parent. Turns out the error was due to a database with a block parent, not another database.

johndgiese (approved):
Looks good

PR #164, opened 2023-12-13
add blocks and tests for the status property

sHermanGriffiths:
# Describe Your Changes
I added the necessary blocks to handle `status` properties and property values.

# How Did You Test It
I added a test for the new property value


sHermanGriffiths:
The lack of handling for this property was getting in the way of Flexview exports, so I added it in there.

bimbashrestha (approved):

PR #167, opened 2024-01-25
Shg/images

sHermanGriffiths:
# Describe Your Changes
Create a plugin that downloads files saved in properties and uses the local versions in the export.

# How Did You Test It
I exported the Case Files database and made sure that the files were downloaded and that the media urls were exported in the properties see screenshots).

# Screenshot
Screenshot 2024-01-25 at 2 13 00 PM
Screenshot 2024-01-25 at 2 13 28 PM
Screenshot 2024-01-25 at 2 13 37 PM


n2y/notion.py
```diff
@@ -466,7 +466,7 @@ def download_file(self, url, page, block_id):
 
     def save_file(self, content, page, extension, block_id):
         block_id_chars = strip_hyphens(block_id)
-        page_title = sanitize_filename(page.title.to_plain_text())
```
johndgiese, line 469: Wait, I don't think this should be slugify here. I think in general we do want underscores.

n2y/plugins/downloadfileproperty.py
```diff
@@ -0,0 +1,28 @@
+import hashlib
+from os import makedirs, path
+from urllib.parse import urlparse
+
+
+from n2y.property_values import FilesPropertyValue
+from n2y.utils import slugify
+
+
+class DownloadFilePropertyValue(FilesPropertyValue):
```
johndgiese, line 10: I think we need a test for this plugin. You'll probably need to extend our plugins end-to-end test ...

johndgiese (changes requested):
See comments

n2y/utils.py
```diff
@@ -172,6 +172,15 @@ def sanitize_filename(filename):
     return s
 
 
+def slugify(string):
+    """Turn a string into a slug."""
+    s = str(string).strip().replace(" ", "-")
```
johndgiese, line 177: We'll want the "slugify" to make the text lower case too. It should work like https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.text.slugify

You could probably even grab the code from Django.

Note slugify is a different function from the sanatize_filename. sanitize_filename should still be used in all of the places it's currently being used in. Also, you don't need to check for "." or ".." in the slugify function.

johndgiese (approved):

PR #172, opened 2024-03-16
Shg/logs

sHermanGriffiths:
# Describe Your Changes
Pass logger around the package as an argument instead of importing it into every module

# How Did You Test It
Update tests to pass the argument and make sure things are running as expected.

# Relevant Task (Rational)
https://www.notion.so/medtech-os/Unrelated-Error-Messages-Being-Printed-e41f32055e0b4992b6b95cbb11a25520?pvs=4

### DIFF
--- a/n2y/audit.py
+++ b/n2y/audit.py
@@ -1,17 +1,17 @@
+import argparse
+import logging
 import os
 import sys
-import logging
-import argparse
 
-from n2y.logger import logger, HANDLER
+from n2y.blocks import LinkToPageBlock
+from n2y.database import Database
+from n2y.errors import UseNextClass
+from n2y.logger import HANDLER
+from n2y.logger import logger as log
+from n2y.mentions import PageMention
 from n2y.notion import Client
 from n2y.page import Page
-from n2y.database import Database
-from n2y.mentions import PageMention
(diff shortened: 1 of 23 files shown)


n2y/audit.py
```diff
@@ -99,7 +100,7 @@ def main(raw_args, access_token):
         },
     }
 
-    client = Client(access_token)
+    client = Client(access_token, logger=logger)
```
russellkan, line 103: Would it be possible to just use `logger=log` here and remove the `logger=log` part of the function def in line 56?

russellkan (approved):
Nice work. One small question but otherwise should be good to merge.

n2y/audit.py
sHermanGriffiths: @russellkan, it definitely would be. I added it on line 56 so that the `main` function can be imported and run elsewhere with the option of running it with a custom logger. Otherwise, there's no point of referencing `logger` on lines 56 or 103 because the default `logger` for the `Client` class is already the package logger (i.e., `n2y.logger.logger`)

P.S. Noice pfp, my guy 🔥

russellkan: Got it. Thanks haha.

PR #174, opened 2024-04-25
Comment out unshared synced block warning

sHermanGriffiths:
# Describe Your Changes
Comment out unshared synced block warning

# How Did You Test It
N/A

### DIFF
--- a/n2y/blocks.py
+++ b/n2y/blocks.py
@@ -868,9 +868,9 @@ class SyncedBlock(Block):
 
     def to_pandoc(self):
         if not self.shared:
-            self.client.logger.warning(
-                "Skipping un-shared synced block (%s)", self.notion_url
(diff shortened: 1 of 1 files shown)


n2y/blocks.py
```diff
@@ -868,9 +868,9 @@ def _get_synced_block_children(self):
 
     def to_pandoc(self):
         if not self.shared:
-            self.client.logger.warning(
```
johndgiese, line 871: Hey @sHermanGriffiths is this the fans warning that we keep seeing printed? If so, I don't think we need comment out the warning. Instead, we should delete it and write a comment explaining why we can't reliably print a warning since there are a lot of false positives etc. etc.

johndgiese (approved):

PR #184, opened 2024-07-01
Shg/api err codes

sHermanGriffiths:
# Describe Your Changes
Modify the `APIErrorCode` to support all error codes supported by the Notion API.

# How Did You Test It
N/A

### DIFF
--- a/n2y/errors.py
+++ b/n2y/errors.py
@@ -1,4 +1,6 @@
-from enum import Enum
+from dataclasses import dataclass
+from enum import Enum, Flag, StrEnum
+from itertools import chain
 
 
 class N2YError(Exception):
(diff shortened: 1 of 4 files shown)


n2y/errors.py
```diff
@@ -67,81 +61,93 @@ def __init__(self, response, message, code) -> None:
...
+    is_retryable: bool
 
-    RestrictedResource = "restricted_resource"
-    """Given the bearer token used, the client doesn't have permission to
-    perform this operation."""
+    def __new__(cls, code: str, is_retryable: bool):
+        obj = str.__new__(cls, code)
+        obj._value_ = code
+        obj.is_retryable = is_retryable
+        cls.RetryableCodes = [i.value for i in cls if i.is_retryable is True]
```
russellkan, line 93: I think `is True` here is unnecessary. And the next line can be `if not i.is_retryable`.

russellkan (approved):

n2y/errors.py
sHermanGriffiths: True, lol

PR #185, opened 2024-07-01
Hash saved files if no block id is provided

sHermanGriffiths:
# Describe Your Changes
Reinstate downloaded file content hashing

# How Did You Test It
N/A

### DIFF
--- a/n2y/notion.py
+++ b/n2y/notion.py
@@ -469,6 +469,10 @@ class Client:
         Preserve the file extension from the URL, but use the
         id of the block followed by an md5 hash.
         """
+        # block_id will be None if the file is not attached to a block
+        # (as is the case for page properties) or one would rather tie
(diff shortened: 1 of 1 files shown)


johndgiese:
Hey @sHermanGriffiths why is this change required?

sHermanGriffiths:
> Hey @sHermanGriffiths why is this change required? Why aren't we able to provide a block id?

Because the images we use the urls from Notion for are Page properties

sHermanGriffiths:
> Hey @sHermanGriffiths why is this change required? Why aren't we able to provide a block id?

@johndgiese Because the images we use the urls from Notion for are Page properties

johndgiese (approved):
Please add a comment explaining the situation when `block_id` would legitametly be None.

PR #190, opened 2024-07-11
Fix: fix broken import for python versions < 3.11

### LINKED ISSUE #188: StrEnum not defined for Python versions < 3.11
The `n2y.errors` module imports `StrEnum`, but this class is not available until 3.11. We should define StrEnum as suggested here: https://docs.python.org/3.10/library/enum.html#others if an import error is encountered.

notmatthancock:
# Describe Your Changes

Catches import error during StrEnum import, and defines the class by subclassing instead.

# How Did You Test It

By `n2y --help` not throwing an error for 3.10 on my machine.

Closes #188

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,6 @@ build/
 dist/
 .vscode
 venv
+.venv
 .tool-versions
 _debug_pkg
--- a/n2y/errors.py
+++ b/n2y/errors.py
@@ -1,4 +1,13 @@
-from enum import StrEnum
+try:
+    from enum import StrEnum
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,5 +6,6 @@ build/
 dist/
 .vscode
 venv
+.venv
```
notmatthancock, line 9: unrelated to the fix, but I added this because it's common to have the virtual environment folder as hidden

sHermanGriffiths (approved):
Good stuff!

PR #191, opened 2024-07-15
Fix APIErrorCode class (did not work as expected)

sHermanGriffiths:
# Describe Your Changes
The previous edits didn't work as expected and raised an error

# How Did You Test It
Write `test_apierrorcode_contains`

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,0 +1,36 @@
+from n2y.errors import APIErrorCode
+
+
+def test_apierrorcode_contains():
+    errors = [
+        ("bad_gateway", True),
+        ("conflict_error", True),
+        ("database_connection_unavailable", True),
+        ("gateway_timeout", True),
+        ("internal_server_error", True),
+        ("invalid_grant", False),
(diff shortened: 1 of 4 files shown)


tests/test_errors.py
```diff
@@ -0,0 +1,36 @@
+from n2y.errors import APIErrorCode
+
+
+def test_apierrorcode_contains():
+    errors = [
```
johndgiese, line 5: Very cool! Checkout pytest fixtures for another way to do this: https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures

johndgiese (commented):
Hey @sHermanGriffiths can you have Bimba review this please? I left one comment

tests/test_errors.py
sHermanGriffiths: Yep yep! We use them in other tests in the notion-qms repo I think. If we expand this module I'll probably end up shifting to that method to avoid definition repetition.

bimbashrestha (approved):

PR #195, opened 2024-07-18
Add internal link hack

notmatthancock:
# Describe Your Changes

This adds a plugin, `internallinks.py`, that fixes broken internal links. Currently if a page has an internal link to a heading contained on that page, the link in the resulting HTML or markdown document points to the original notion source. This PR patches the links to instead be the fragment of the ID of the header.

# How Did You Test It

I added unit tests in this PR, and I checked the `n2y` CLI against a page on the test Notion workspace.

### DIFF
--- a/n2y/plugins/internallinks.py
+++ b/n2y/plugins/internallinks.py
@@ -3,6 +3,7 @@ import uuid
 import urllib.parse
 
 from n2y.blocks import Block, ChildPageBlock
+from n2y.errors import UseNextClass
 from n2y.logger import logger
 from n2y.page import Page
 from n2y.rich_text import TextRichText
@@ -42,16 +43,20 @@ def find_target_block(page: Page, target_id: str) -> Block:
 
 
 class NotionInternalLink(TextRichText):
-    """ This plugin fixes internal links to headers on the same page.
+    """This plugin fixes internal links to headers on the same page.
 
     Instead of leaving the link to point to the original notion source page,
     it substitutes the link to instead be the ID of the header.
     """
 
-    def to_pandoc(self):
-        if not is_internal_link(self.href, self.block.page.notion_id):
-            return super().to_pandoc()
+    def __init__(self, client, notion_data, block=None):
+        super().__init__(client, notion_data, block)
+        if block is None or not is_internal_link(
+            self.href, self.block.page.notion_id
+        ):
+            raise UseNextClass
 
+    def to_pandoc(self):
         target_id = get_notion_id_from_href(self.href)
         if target_id is None:
             logger.warning(
--- a/n2y/plugins/internallinks.py
+++ b/n2y/plugins/internallinks.py
@@ -28,11 +28,10 @@ def is_internal_link(href: typing.Optional[str], notion_id: str) -> bool:
 
 
 def find_target_block(page: Page, target_id: str) -> Block:
-    try:
-        page_block: ChildPageBlock = page.block
-    except AttributeError:
-        logger.error(f"page block")
-        raise
+    page_block: ChildPageBlock = page.block
+    if page_block is None:
+        raise ValueError("Page.block is None")
+
     try:
         return next(
             child for child in page_block.children if child.notion_id == target_id
@@ -44,7 +43,7 @@ def find_target_block(page: Page, target_id: str) -> Block:
 
 class NotionInternalLink(TextRichText):
     """ This plugin fixes internal links to headers on the same page.
-    
+
     Instead of leaving the link to point to the original notion source page,
     it substitutes the link to instead be the ID of the header.
     """
--- a/n2y/plugins/internallinks.py
+++ b/n2y/plugins/internallinks.py
@@ -2,10 +2,9 @@ import typing
 import uuid
 import urllib.parse
 
-from n2y.blocks import Block, ChildPageBlock
+from n2y.blocks import Block
 from n2y.errors import UseNextClass
 from n2y.logger import logger
-from n2y.page import Page
 from n2y.rich_text import TextRichText
 from n2y.utils import header_id_from_text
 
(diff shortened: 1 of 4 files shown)


n2y/plugins/internallinks.py
```diff
@@ -0,0 +1,68 @@
...
+class NotionInternalLink(TextRichText):
+    """ This plugin fixes internal links to headers on the same page.
+
+    Instead of leaving the link to point to the original notion source page,
+    it substitutes the link to instead be the ID of the header.
+    """
+
+    def to_pandoc(self):
+        if not is_internal_link(self.href, self.block.page.notion_id):
+            return super().to_pandoc()
```
johndgiese, line 53: Is there a reason why we couldn't check whether it's an internal link in the `__init__` and then raise a `UseNextClass` exception if it's not? This would allow people to have alternate plugins for the `TextRichText` class that could apply for non-internal links.

tests/test_plugin_internallinks.py
```diff
@@ -0,0 +1,112 @@
+from unittest.mock import Mock, patch
```
johndgiese, line 1: Nice job following the existing patterns with the testing. This is all great, nice work!

johndgiese (approved):
Very nice work @notmatthancock ! I had one comment that should be addressed most likely. Feel free to merge otherwise.

n2y/plugins/internallinks.py
notmatthancock: That's a good idea. It breaks the tests though, but it's probably because of how the test constructs the test page and its contents. I'll have to look into it.

n2y/plugins/internallinks.py
```diff
@@ -0,0 +1,68 @@
...
+import urllib.parse
+
+from n2y.blocks import Block, ChildPageBlock
+from n2y.logger import logger
+from n2y.page import Page
+from n2y.rich_text import TextRichText
+from n2y.utils import header_id_from_text
+
+
+def get_notion_id_from_href(href: str) -> typing.Optional[str]:
```
sHermanGriffiths, line 12: `... -> typing.Optional[str]:` can be simplified to `...-> str | None` if you want. It's a preference choice, but just making sure you're aware.

n2y/plugins/internallinks.py
```diff
@@ -0,0 +1,68 @@
...
+    return href is not None and href.startswith(f"/{notion_id.replace('-', '')}")
+
+
+def find_target_block(page: Page, target_id: str) -> Block:
+    page_block: ChildPageBlock = page.block
+    if page_block is None:
+        raise ValueError("Page.block is None")
+
+    try:
+        return next(
```
sHermanGriffiths, line 36: What if the `target_id` is to a child of one of the `ChildPageBlock's` children? Or the child of one of its children's children? This might need to be recursive logic.

sHermanGriffiths (commented):
Nice! It looks good, I just a couple things you might want to consider

n2y/plugins/internallinks.py
notmatthancock: That syntax was introduced in 3.10 and the setup.py file indicates n2y might be intended to be compatible with earlier Python versions.

n2y/plugins/internallinks.py
notmatthancock: Good catch. Can you give some concrete examples where this would occur so I can understand the usecase more (I don't want to over generalize too soon)? Like a heading inside of a ___ ?

sHermanGriffiths: So, the `ListItemBlock` objects would be a good example as they are really just a serialization container for rich text arrays. They may also have `TableBlock`, `ParagraphBlock`, or `ListItemBlock` children of their own. Speaking of, `ParagraphBlock` objects can be children of each other. `TableBlock` objects have `RowBlock` children. `TableOfContentsBlock` objects have `TableOfContentsItemBlock`. `ToggleBlock` objects work similar to `ListItemBlock` objects, I believe. Lastly, I think `ColumnListBlock` object's children are `ColumnBlock` objects, which will be able to have children of all the same type as the `ListItemBlock` objects.

n2y/plugins/internallinks.py
sHermanGriffiths: Gotcha. That needs to get updated, actually. n2y is supported for up to 3.11 not just 3.9

sHermanGriffiths:
@notmatthancock Can you also note your changes in the changelog under `0.10.3` to track all of our changes once you upversion?

n2y/plugins/internallinks.py
johndgiese: @sHermanGriffiths note that n2y is used with the website so if we use features that require a new version of python we may force the website to update their python too if that makes sense.

sHermanGriffiths: Good point!

n2y/plugins/internallinks.py
sHermanGriffiths: @notmatthancock, about the concern you raised during our call, I would note that the block may be `None` whether `is_internal_link` is called in `to_pandoc` or `__init__.` To mitigate this risk, I'd write something like the following for `__init__`:
```py
    def __init__(self, client, notion_data, block):
        super().__init__(client, notion_data, block)
        if not block or not is_internal_link(self.href, self.block.page.notion_id):
            raise UseNextClass
```

notmatthancock: https://github.com/innolitics/n2y/pull/195/commits/c56fdd714d00b67d643a3cf5d16d4849b88b15c3

n2y/plugins/internallinks.py
notmatthancock: Updated here: https://github.com/innolitics/n2y/pull/195/commits/08f29d0a604cda9aa95defb310e08e8d2f4b3da4

notmatthancock:
> Can you also note your changes in the changelog under 0.10.3 to track all of our changes once we upversion?

Done.

PR #199, opened 2024-09-24
Prevent image collisions by using the full image id

sHermanGriffiths:
# Describe Your Changes
Notion's ID system gives images or blocks on the same page similar IDs (e.g. [this image block](https://www.notion.so/DOC-0083-Validation-Performance-Testing-Report-7b66917ff7c6435e91f1fd678bbcec81?pvs=4#1038bd408bfb8050ba13ceed672774db), [this image block](https://www.notion.so/DOC-0083-Validation-Performance-Testing-Report-7b66917ff7c6435e91f1fd678bbcec81?pvs=4#1038bd408bfb807593eee3783c488d7a), and many of the following image blocks on that page in the Galileo workspace). This led to naming collisions, as we only used the first 11 characters of the block ID to identify the files. I remember the request for this convention, but I don't remember the reasoning that inspired it. In any case, I'm assuming it was simply to have shorter file names, and I believe that extending the file names to include the whole notion block ID won't cause any harm to the systems we have in place.

# How Did You Test It
Ensure tests are still passing

### DIFF
--- a/n2y/notion.py
+++ b/n2y/notion.py
@@ -476,7 +476,7 @@ class Client:
     def save_file(self, content, page, extension, block_id):
         id_chars = strip_hyphens(block_id)
         page_title = sanitize_filename(page.title.to_plain_text())
-        relative_filepath = f"{page_title}-{id_chars[:11]}{extension}"
+        relative_filepath = f"{page_title}-{id_chars}{extension}"
         full_filepath = path.join(self.media_root, relative_filepath)
         makedirs(self.media_root, exist_ok=True)
         with open(full_filepath, "wb") as temp_file:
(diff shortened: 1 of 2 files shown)


ReeceStevens (approved):
This looks fine to me. I went diving into the git blame and I can't find a clear reason for why this was added either. Generally speaking, the only reasons I can think of for truncating the ID would be:

1. Visual appearance of the downloaded file name
2. Windows file path length restrictions

I don't think these notion IDs are long enough to really worry about (2), so I'd say your approach seems fine to me.

sHermanGriffiths:
> This looks fine to me. I went diving into the git blame and I can't find a clear reason for why this was added either. Generally speaking, the only reasons I can think of for truncating the ID would be:
> 
> 1. Visual appearance of the downloaded file name
> 2. Windows file path length restrictions
> 
> I don't think these notion IDs are long enough to really worry about (2), so I'd say your approach seems fine to me.

My inclination was that it was 1. as well

PR #200, opened 2024-11-09
Get rid of the global logger

sHermanGriffiths:
# Describe Your Changes
The use of the global logger resulted in the configuration of the logger in any project that uses the `n2y` package. This is obviously not the desired effect, as we want `n2y` users to be able to use our package and configure their logging system the way they would like.

# How Did You Test It
I loaded n2y in a project and ensured that the project's logging configuration was used and not the n2y package's.


n2y/plugins/internallinks.py
```diff
@@ -64,22 +63,22 @@ class NotionInternalLink(TextRichText):
...
-            self.href, self.block.page.notion_id
-        ):
+        if block is None or not is_internal_link(self.href, self.block.page.notion_id):
             raise UseNextClass
 
     def to_pandoc(self):
         target_id = get_notion_id_from_href(self.href)
         if target_id is None:
-            logger.warning(
+            self.client.logger.warning(
```
ReeceStevens, line 72: I think it might make sense to stick with using a module-level logger here, rather than pulling in the logger from the `self.client` property. I might be missing context though-- is there a reason you took this approach instead?

ReeceStevens (changes requested):
Looks good to me, just one minor comment. I always have to remind myself how Python logging works in situations like these!

n2y/plugins/internallinks.py
sHermanGriffiths: Yes, we allow users to pass a logger to the `Client`, so we want all `Block`, `Page`, and `Database` objects to use `self.client.logger` to utilize any logger that may be passed.

ReeceStevens: Ahh, I see. Thanks for that clarification!

ReeceStevens (approved):
LGTM :+1:

PR #205, opened 2025-08-20
Implement changes for allowing link_mention parsing from notion API

slohat-enestit:
… to Pandoc
# Describe Your Changes

Implemented support for parsing link_mention objects from the Notion API into Pandoc.

Added a helper method get_user_pages in n2y/notion.py to retrieve user pages.

Added an end-to-end test test_link_mention_with_page in tests/test_rich_text_mentions.py.

#How Did You Test It

Created unit tests with mock data to validate Pandoc AST, Markdown, and plain-text conversions.

Added an end-to-end integration test against a real Notion page to ensure link_mention is parsed and exported correctly.

### DIFF
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -510,3 +510,20 @@ Yakkity yakkity yakkity yak
             "headers to maintain Markdown spec"
         )
         assert any(expected_blurb in m for m in captured_messages)
+
+
+def test_link_mention_with_page(tmpdir):
+    """
+    Test that a Notion page containing an inline link mention is exported correctly.
+    The page can be seen here:
+    https://fresh-pencil-9f3.notion.site/Simple-Link-Mention-Test-2556bc16d85d80a79f05f39a1932d0a2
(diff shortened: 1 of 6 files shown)


tests/test_end_to_end.py
```diff
@@ -510,3 +510,20 @@ def test_simple_table_headers(caplog, tmp_path):
             "headers to maintain Markdown spec"
         )
         assert any(expected_blurb in m for m in captured_messages)
+
+
+def test_link_mention_with_page(tmpdir):
```
johndgiese, line 515: Very nice work creating a test! I would suggest that in this case we actually update the `test_all_blocks_page_to_markdown` to help keep the testing times quicker, since we already have a big test just for this page. With the unit tests, we definitely recommend small/quick tests, but with the end-to-end tests (since they're slow) sometimes we sacrifice having small atomic tests to make the test suite run faster.

johndgiese (changes requested):
This mostly looks good! I had a few small comments before merging.

johndgiese (approved):
Nice work!

